PR #8, opened 2025-08-29
add optional chaining as metadata can be undefined

Leonidaz:
ran into an issue when calling a child component from App, this fixes it.  It looks like optional chaining was already above this line with the usage of `metadata`


trueadm (approved):
lgtm

PR #16, opened 2025-08-31
Fixed reactive object property decrement not working issue

SIB61:
### Issue: 
-    with reactive obj prop { $count : 1 }, obj.$count -- doesn't work. Instead it increase the count. 

### Fix: 
-    decrement the value instead of always incrementing it.


trueadm:
Mind formatting it properly please? Thanks for the change!

SIB61:
Updated the PR with proper formatting.

trueadm (approved):
Thank you

PR #28, opened 2025-09-01
perf,chore(vite): use hook filters

### LINKED ISSUE #22: Add plugin hook filters
Hey 👋 

Congrats on the release of RIpple! I've noticed that the [transform hook](https://github.com/trueadm/ripple/blob/5139a61eb353e2d78e60f8ba81e164c3ccc0bea0/packages/vite-plugin-ripple/src/index.js#L54-L56) in Ripple's Vite plugin has no hook filter yet, which would improve performance with `rolldown-vite` (and the next Vite major version).

See also:

* https://vite.dev/guide/rolldown.html#hook-filter-feature
* https://github.com/e18e/ecosystem-issues/issues/171

JakeIsMeh:
Needed to bump the vite version as it was too old.
Also tested with rolldown-vite.

Closes #22

### DIFF
--- a/packages/vite-plugin-ripple/src/index.js
+++ b/packages/vite-plugin-ripple/src/index.js
@@ -51,20 +51,23 @@ export function ripple(inlineOptions) {
 				}
 			},
 
-			async transform(code, id, opts) {
-				const filename = id.replace(root, '');
-				if (!filename.endsWith('.ripple')) return;
+			transform: {
+				filter: { id: /\.ripple$/ },
 
-				const { js, css } = await compile(code, filename, id);
+				async handler(code, id, opts) {
+					const filename = id.replace(root, '');
 
-				if (css !== '') {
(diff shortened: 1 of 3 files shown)


trueadm:
This seems to break the tests. :/

JakeIsMeh:
Seems like vitest itself didn't support hook filters.

JakeIsMeh:
Since we're bumping deps in this PR, do you want me to do the same for all packages in the repo? (e.g. playground also has vitest but @^2 rather than @^3)

JakeIsMeh:
(I think using pnpm catalogs may be better, but it might be out of scope for this issue.)

trueadm:
Go for it :) thank you!

JakeIsMeh:
I'll migrate to catalogs in another PR. feel free to merge this first.

PR #29, opened 2025-09-01
trim code before compiling

### LINKED ISSUE #17: variable declaration in multiple scopes causes error - Unterminated JSX contents
declaring variables in "html"/"jsx" blocks causes an `Unterminated JSX contents` error.  Moving the second declaration out of the local scope and to the top  of the component works but it seems that the intention is for local scopes to work as one scope cannot see the others variables.

```jsx
export component App() {
  <div>
    let obj = { $count: 0};

    <span>{"parent: "} {obj.$count}</span>
  </div>

  <div>
    let b = { $count: 0 }; // <----- this is the issue

    <button onClick={() => b.$count--}>{"-"}</button>
    <span>{b.$count}</span>
    <button onClick={() => b.$count++}>{"+"}</button>
  </div>
}
```

if I move out the second declaration to the top of the component it works:

```jsx
export component App() {
    let b = { $count: 0 }; // <----- moving to the top of the component from the second scope works

  <div>
    let obj = { $count: 0};

    <span>{"parent: "} {obj.$count}</span>
  </div>

  <div>
    <button onClick={() => b.$count--}>{"-"}</button>
    <span>{b.$count}</span>
    <button onClick={() => b.$count++}>{"+"}</button>
  </div>
}
```

Leonidaz:
fixes #17

### DIFF
--- a/packages/vite-plugin-ripple/src/index.js
+++ b/packages/vite-plugin-ripple/src/index.js
@@ -55,7 +55,7 @@ export function ripple(inlineOptions) {
 				const filename = id.replace(root, '');
 				if (!filename.endsWith('.ripple')) return;
 
-				const { js, css } = await compile(code, filename, id);
+				const { js, css } = await compile(code.trim(), filename, id);
 
 				if (css !== '') {
 					const cssId = createVirtualImportId(filename, root, 'style');


trueadm:
Thanks but this will likely mess up sourcemaps and the language server, so I'll try and fix it properly :)

PR #38, opened 2025-09-02
update basic template deps

Leonidaz:
it was breaking on install otherwise.

also added node 20 as the minimum support.


trueadm:
Thanks

Issue #44, opened 2025-09-02
Extension not working on Windows – ESM loader error


kravallee:
Hi,
the extension doesn’t seem to work on Windows for me. I’m not sure if this is an error on my side or an actual issue with the extension.

When I open a .ripple file, I get the following error in vscode:

```pgsql
[Error - 11:27:02 AM] Restarting server failed
  Message: Request initialize failed with message: Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
  Code: -32603
```

System info:
- OS: Windows 11
- VS Code version: 1.103.2
- Extension version: 0.0.5

alesvaupotic:
Confirming on Win 10, Win 11, VSCode 1.103.2

trueadm:
Thanks. I don’t have a Windows machine to test but let me dig into it

PR #45, opened 2025-09-02
fix: extension crashing on windows

vyntrix:
Changed the server connection from lsp.TransportKind.ipc to lsp.TransportKind.stdio and updated the server to use pathToFileURL when dynamically importing ripple_path.

Closes: #44

Notes:
- Using stdio is a bit slower than IPC, but it’s much more reliable with ESM on Windows.
- A faster IPC-based solution might be possible in the future, but this fixes the immediate crash.


trueadm (approved):
Thank you!

PR #46, opened 2025-09-02
add playground

aecsar:


trueadm:
Can we remove the HTML and CSS tabs at the top?

aecsar:
Let me double check the livecodes docs

trueadm:
Thanks!

trueadm:
I updated it so the playground output looks a little nicer, but it has a flash of white. What am I missing?

trueadm:
We also need to get syntax highlighting working for the `<style>` block

aecsar:
Ohh I think it's related to the original page background before the editor is loaded (on dark mode). I'll update that and load a corresponding background based on the current theme

PR #47, opened 2025-09-02
add array.from support

Leonidaz:
`array.from` has the same signature as `Array.from` and can handle any array-like objects as well arrays.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,19 @@ You'll need to import the array function from Ripple:
 import { array } from 'ripple';
 
 const arr = array(1, 2, 3);
+
+// or alternatively
+const arr = array(...[1, 2, 3]);
+```
+
+If you already have an array or array-like object, you can create a Ripple array by calling `array.from`.  It has the same signature as the standard `Array.from` function:
+
+```js
+import { array } from 'ripple';
+
+const existingArr = [1, 2, 3];
+
+const arr = array.from(existingArr);
 ```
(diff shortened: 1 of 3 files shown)


trueadm:
I think we probably want to expose `RippleArray` so that people can subclass it and `instanceof` it, but also so we can put `from` on it?

leonidaz:
do you want to get rid of `array` function then and just use `new RippleArray(1,2,3)` and `RippleArray.from([1,2,3])`?

trueadm:
What do you think? It's probably the better approach once we have the other data structures like Map and Set etc.

leonidaz:
In addition to subclassing and `instanceOf`, I think it would more intuitive and more obvious to use classes: `RippleArray`, `RippleSet`, and `RippleMap`.  I'm personally fine with functions but it seems like it would be a topic for constant arguments and a time drain 😄

trueadm:
Let's go that direction then! :)

leonidaz:
cool, i'll make the changes.

ryanatkn:
I didn't want to add opinionated noise on other issues, but I really like `RippleArray` over `array` for consistency rather than the one special case. It also matches Intellisense this way. A special syntax for convenience wouldn't cause the same downsides as `array` does IMO. (edit: I hadnt seen the recent discussion before writing this, I'll contribute there if I feel I have something to add)

PR #51, opened 2025-09-03
replace array with RippleArray

Leonidaz:
- also adds RippleArray.from and RippleArray.of static methods

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,16 +181,23 @@ Now `$count` will only reactively create its value on initialization.
 
 When creating state that contains arrays you should use the reactive alternative for arrays that Ripple provides.
 
-You'll need to import the array function from Ripple:
+You'll need to import the `RippleArray` class from Ripple.  It extends the standard JS `Array` class, and supports all of its methods and properties.
 
 ```js
-import { array } from 'ripple';
+import { RippleArray } from 'ripple';
 
-const arr = array(1, 2, 3);
+// using the new constructor
+const arr = new RippleArray(1, 2, 3);
+
+// using static from method
+const arr = Ripple.from([1, 2, 3]);
+
+// using static of method
+const arr = Ripple.of(1, 2, 3);
 ```
 
-The `array` is a reactive Ripple array, and that means you can access properties normally using numeric index. However,
-accessing the `length` property of a Ripple reactive array will be not be reactive, instead you should use `$length`.
+The `RippleArray` is a reactive array, and that means you can access properties normally using numeric index. However,
+accessing the `length` property of a `RippleArray` will be not be reactive, instead you should use `$length`.
 
 ### Effects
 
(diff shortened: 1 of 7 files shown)


leonidaz:
not sure why the test failed as I haven't change, i'll check tomorrow once I fully set up the repo env.

sreekar339339:
Is there a restriction in calling the class `$Array`? It's less characters than `RippleArray` and the $ sign indicates the reactive nature of it, consistent with the naming convention of Ripple's philosophy.

leonidaz:
It's not my call but it's just a class or a constructor and in my opinion what matters is whether the variable that gets the assignment are reactive or not.

What's the difference of calling it `$Array` or `RippleArray` if they both signify that the items inside the array are reactive.  It's immediately recognizable.  Nothing else is going to be called RippleArray that would mean otherwise, unless one creates something custom, so it's easy to learn.  If you want you can create your own `export const $Array = RippleArray;` and import it instead  if it makes it more readable to you _(although not sure if currently ripple can handle the parsing)_

I think prefixing classes with a `$` just pollutes the code and waters down the concept of reactive variables and props.

Anyway, I think @trueadm had some other ideas on our other thread, to which I had some questions: https://github.com/trueadm/ripple/issues/30#issuecomment-3246828143

So I think we should wait for him to figure out where he wants to go with that.

trueadm:
The PR needs to update the tests that are failing. As for `$Array`, that's not something we should do. `RippleArray` is fine as it is, if we put a `$` prefix on it, and do `$Array.from` then the compiler will attempt to treat it like a reactive binding, which it's not.

leonidaz:
> The PR needs to update the tests that are failing.

@trueadm the test are failing because there is some compilation that was happening when calling `array()`

e.g. this
`const items = array('Item 1', 'Item 2', 'Item 3');`

gets compiled to this:
`const items = $.with_scope(__block, () => array('Item 1', 'Item 2', 'Item 3'));`

where as for calling the RippleArray constructor, the compiled code remains:
`const items = new RippleArray('Item 1', 'Item 2', 'Item 3');`

however, if I use the static `RippleArray.of` or `RippleArray.from`, since these are function calls, the compiler properly recognizes it, via the `CallExpression`:

`const items = $.with_scope(__block, () => RippleArray.of('Item 1', 'Item 2', 'Item 3'));`

So, it seems that `new RippleArray()` construct would be a special case, and since I'm not (at this point) familiar with the compiler code, I'm not sure how to implement it.  Also, it would be a special case since other class constructors should not be wrapped with `$.with_scope`.

I can change the tests for now to use the static RippleArray methods so they pass, but I think you'd agree that it's best that you take over the branch and decide how to parse the constructor properly.

What do you think?

trueadm:
That's a bug that I need to fix, `new` expressions should also get wrapped, let me sort that out.

Update: pushed a fix.

PR #57, opened 2025-09-03
fix grammar regex: remove variable-length lookbehind

vyntrix:
Replaced three variable-length lookbehinds with fixed-length lookaheads in ripple.tmLanguage.json to fix grammar loading errors.

this should hopefully fix this error when building grammar for linguist:

```shell
1 errors found in new grammar 'repository `vendor/grammars/ripple` (from https://github.com/trueadm/ripple.git)':
- Invalid regex in grammar: `source.ripple` (in `packages/ripple-vscode-plugin/syntaxes/ripple.tmLanguage.json`) contains a malformed regex (regex "`(?!(?<![_$[:alnum:]])(?:(?<=\.\.`...": lookbehind assertion is not fixed length (at offset 731))
```

Notes:
- Extension should work like before (tested)


leonidaz:
should we bump up the `ripple-vscode-plugin` or you intentionally left it unpublished for now?

trueadm:
I just published it

PR #70, opened 2025-09-04
playground improvements

hatemhosny:
This PR adds some minor improvements to the playground.

- Pins the SDK version to avoid later breaking changes.
- By default, some styles are applied for embedded playgrounds (e.g. border, border-radius, etc.), which may not be ideally when the playground fills the page. [Default styles were disabled](https://livecodes.io/docs/sdk/js-ts#default-styles) and custom styles were added.
- Some other corrections for the use of the [SDK](https://livecodes.io/docs/sdk/js-ts).

I'm happy to continue maintaining the playground in the website. Please let me know (in [this issue](https://github.com/trueadm/ripple/issues/9)) if you need any modifications.

### DIFF
--- a/website/playground.html
+++ b/website/playground.html
@@ -4,16 +4,29 @@
 		<meta charset="UTF-8" />
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<title>Ripple playground</title>
+		<style>
+			html,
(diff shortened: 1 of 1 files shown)


trueadm (approved):
Thanks :)

PR #74, opened 2025-09-05
feat: RippleSet

leonidaz:
- added RippleSet support, once approved, will work on RippleMap
- added some basic tests
- added a new section to ReadMe

- here's some code for the playground if you want to give it a whirl in the ui:
```jsx
import { RippleSet } from 'ripple';

export component App() {
  const mySet = new RippleSet([1, 2, 3]);
  const other = new RippleSet([1, 2, 3, 4, 5, 6]);
  const $val1 = mySet.has(4);
  let $exists = false;

  console.log(JSON.stringify(mySet));

  <div>
    <button onClick={() => {
      const toAdd = mySet.size + 1;
      $exists = mySet.has(toAdd);

      if (!$exists) {
        mySet.add(toAdd);
      }
    }}>{"Add Item"}</button>
  </div>

  if ($exists) {
    <div class="warning">{"already exists and will not be added."}</div>
  }

  <div>{"$val1 is present: "}{$val1}</div>

  <div>{"mySet is isSubsetOf: "}{mySet.isSubsetOf(other)}</div>

  <pre>{"mySet: "}{JSON.stringify(mySet)}</pre>

  if (mySet.$size === 0) {
    <div>{"Set is empty"}</div>
  }

  if (mySet.has(4)) {
    <div>{"Set has 4"}</div>
  } else {
    <div>{"Set does not have 4"}</div>
  }

  let $values = mySet.values();

  for (const item of $values) {
    <div>
      <button onClick={() => mySet.delete(item)}>
        {"Remove "} {item}
      </button>
    </div>
  }

  <style>
      button {
        margin: 8px;
        padding: 8px;
        font-size: 16px;
      }

      .warning {
        color: red;
        font-weight: bold;
      }
  </style>
}

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ component Button(props: { text: string, onClick: () => void }) {
 }
 
 // Usage
-component App() {
+export component App() {
   <Button text="Click me" onClick={() => console.log("Clicked!")} />
 }
 ```
@@ -175,7 +175,7 @@ Now `$count` will only reactively create its value on initialization.
 
 > Note: you cannot define reactive variables in module/global scope, they have to be created on access from an active component
 
-#### Arrays
+#### Reactive Arrays
 
 When creating state that contains arrays you should use the reactive alternative for arrays that Ripple provides.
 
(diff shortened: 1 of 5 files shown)


leonidaz:
interesting, the test is passing locally...

kravallee:
> interesting, the test is passing locally...

i think it's some prettier thing right now

trueadm:
mind rebasing main. it seems like something is up, it could be the node version on CI vs yours too?

trueadm:
> i think it's some prettier thing right now

Prettier tests are passing in CI on this branch though?

kravallee:
> Prettier tests are passing in CI on this branch though?

sorry my fault didn't read it correctly 😅

leonidaz:
yeah, it's failing with node 18 and 20, I was using 22.

I'll dig into this.

leonidaz:
lol, some of the Set methods were only added last year... no problem I'll make them optional.

leonidaz:
or should we shim them?

trueadm:
It looks like you based this on the Svelte version? I guess they must run node 22 in CI

leonidaz:
> It looks like you based this on the Svelte version? I guess they must run node 22 in CI

Yeah, I worked with both svelte and your array implementation.  but Svelte doesn't support these new methods so it wouldn't be an issue for them anyway.

trueadm:
What if we check they're there before assigning them?

leonidaz:
@trueadm 

they're easily shimmable if we want or I can just add them if they're available, what do you think?

E.g. this is a shim for Set.union

```js
if (!Set.prototype.union) {
  Set.prototype.union = function(other) {
    if (!(other instanceof Set)) {
      throw new TypeError('Argument must be a Set');
    }
    const result = new Set(this);
    for (const item of other) {
      result.add(item);
    }
    return result;
  };
}
```



> What if we check they're there before assigning them?

yeah, that's very easy too, just thought maybe we want to be more "advanced" and provide some sugar.

trueadm:
Let’s just keep it simple I think

leonidaz:
ok, all good now.

trueadm:
Love it!

PR #75, opened 2025-09-05
fix `<code>` legibility in the playground

nolanlawson:
This might just be a Linux issue (although it repros in both Chrome and Firefox), but the `<code>` in the playground is unreadable for me. Before and after:

Screenshot from 2025-09-05 09-55-44
Screenshot from 2025-09-05 09-55-49

Very exciting project BTW!


trueadm (approved):

trueadm:
Thank you!

PR #76, opened 2025-09-05
pin ripple version in playground

hatemhosny:
The latest ripple version (v0.2.17) broke the playground: https://www.ripplejs.com/playground

Pinning ripple version in playground to v0.2.16 till this is fixed.

### DIFF
--- a/packages/ripple/tests/basic.test.ripple
+++ b/packages/ripple/tests/basic.test.ripple
@@ -344,4 +344,476 @@ describe('basic', () => {
 
 		expect(container.querySelector('.count').textContent).toBe('0');
 	});
+
+	it('renders with different event types', () => {
+		component Basic() {
+			let $focusCount = 0;
+			let $clickCount = 0;
+
+			<button 
+				onFocus={() => $focusCount++}
+				onClick={() => $clickCount++}
+			>{'Test Button'}</button>
+			<div class='focus-count'>{$focusCount}</div>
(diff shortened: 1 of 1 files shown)


trueadm:
What's broken?

hatemhosny:
image

After updating to new versions, I start getting such errors.
However, when I pin it to any version, it works correctly.
So it seems that the version "latest" causes caching of files across different versions (either at browser or CDN levels), causing incompatibility.

I usually pin versions of dependencies, but in this case due to the rapid development and the frequent publishing we are leaving it to latest version.

What do you think?

leonidaz:
there is quite a bit of caching in the browser.  Setting the `disable cache` in the Network tab in the dev tools helped.  But I also had to clear site data in Application quite a bit.  

This was when I was using the basic template.  it looked like the .vite bundle gets cached.  Anyways, I didn't look into it too much.

hatemhosny:
One option is actually checking for versions before loading the playground
https://data.jsdelivr.com/v1/packages/npm/ripple

This can also allow us to prefill a dropdown list of versions that allows users to [change Ripple version](https://github.com/trueadm/ripple/issues/9#issuecomment-3255863645) used by the playground during runtime.

demo: https://livecodes.io/?x=id/mmiuf6iq6ac

leonidaz:
that would be ideal actually! 👍

PR #77, opened 2025-09-05
fix: set missing scope on has

leonidaz:
oops.

we should probably start jsdoc'ing code sometime soon.


trueadm (approved):

trueadm:
A test should cover this API too?

PR #79, opened 2025-09-05
test: added more basic tests

kravallee:
Added some more basic tests mainly about event handling and component rendering

### DIFF
--- a/packages/ripple/tests/basic.test.ripple
+++ b/packages/ripple/tests/basic.test.ripple
@@ -344,4 +344,476 @@ describe('basic', () => {
 
 		expect(container.querySelector('.count').textContent).toBe('0');
 	});
+
+	it('renders with different event types', () => {
+		component Basic() {
+			let $focusCount = 0;
+			let $clickCount = 0;
+
+			<button 
+				onFocus={() => $focusCount++}
+				onClick={() => $clickCount++}
+			>{'Test Button'}</button>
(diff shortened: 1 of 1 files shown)


trueadm (approved):

trueadm:
Thank you!

packages/ripple/tests/basic.test.ripple
```diff
@@ -344,4 +344,476 @@ describe('basic', () => {
...
+				$hasRef = !!node;
+			}}>{'Element with ref'}</div>
+			<div class='has-ref'>{$hasRef ? 'Has ref' : 'No ref'}</div>
+		}
+
+		render(Basic);
+
+		const hasRefDiv = container.querySelector('.has-ref');
+		expect(container.querySelector('.test-element')).toBeTruthy();
+	});
```
leonidaz, line 441: i was sure ref's were removed.  

I'm actually not sure what this test is testing other then the presence of the `.test-element`

it's the `textContent` of `hasRefDiv` that could be conditional but it's never tested and since refs are gone, it would fail.

trueadm: They were removed. We should refactor this test tbh. Sorry I missed it in review

kravallee: sorry I messed up I thought refs are still a thing

PR #83, opened 2025-09-06
use latest ripple version in playground

hatemhosny:
This is a follow up for #76 

It checks latest published version from https://data.jsdelivr.com/v1/packages/npm/ripple
and uses that in the playground to avoid the cache problem outlined in #76

In addition, the css unit `dvh` is used instead of `vh` to avoid hiding the lower part (including console & compiled code viewer) in mobile.

image

### DIFF
--- a/website/playground.html
+++ b/website/playground.html
@@ -25,6 +25,7 @@
 			createPlayground('#container', {
 				appUrl: 'https://ripple.livecodes.pages.dev',
 				config: {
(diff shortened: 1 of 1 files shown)


hatemhosny:
I did not implement version selector as discussed in https://github.com/trueadm/ripple/pull/76#issuecomment-3260016051 for now till the docs website is ready to maintain a uniform UI.

trueadm (approved):

PR #100, opened 2025-09-07
feat: add slot syntax with Component.$$slot annotations

nihgwu:
# A proposal to https://github.com/trueadm/ripple/issues/50, created by Claude Code

## Summary

Adds a new slot syntax using `Component.$$slot = 'slotName'` annotations that provides a cleaner alternative to inline component definitions within JSX elements. To eliminate duplication of inline slot notations and enable better component composition:

in ui library:

```jsx
component Button({ $leadingIcon, $children, $trailingIcon ]) {
  <button>
    <span class='button-leading-icon'><$leadingIcon /></span>
    <$children />
    <span class='button-trailing-icon'><$trailingIcon /></span>
  </button>
}

component LeadingIcon({ icon }) { <svg>{icon}</svg> }
Header.$$slot = 'leadingIcon'

component TrailingIcon({ icon }) { <svg>{icon}</svg> }
Footer.$$slot = 'trailingIcon'
```

in app:

```jsx
<Button>
  <LeadingIcon icon={iconA} />
  {"Label"}
  <TrailingIcon icon={iconB} />
</Button>
```

without the slot notation, user has to do

```jsx
<Button>
  component leadingIcon() { <LeadingIcon icon={iconA} /> }
  {"Label"}
  component trailingIcon() { <TrailingIcon icon={iconB} /> }
</Button>
```

**Imagine you have many buttons on a single page, how it will be**

---

Some people would say, why not just

```
<Button
  $leadingIcon={<LeadingIcon icon={iconA} />}
  $trailingIcon={<TrailingIcon icon={iconB} />}
>
  {'Label"}
</Button>
```

Sure that works without changing anything, but this is not how WC's template/slot works
Personally I don't like this **fat props thin children** style, and nowadays the composition style is prevailing, thought it makes Accessibility support more complex(context) and style challenging(presence detection)

### DIFF
--- a/packages/ripple/src/compiler/phases/2-analyze/index.js
+++ b/packages/ripple/src/compiler/phases/2-analyze/index.js
@@ -321,12 +321,7 @@ const visitors = {
 							binding.transform = {
 								read: (_) => b.call('$.get_computed', path.node),
 								assign: (node, value) => {
-									return b.call(
-										'$.set',
-										path.node,
-										value,
-										b.id('__block'),
-									);
+									return b.call('$.set', path.node, value, b.id('__block'));
 								},
 								update: (_) =>
 									b.call(
@@ -537,6 +532,25 @@ const visitors = {
 
 		context.next();
 	},
+
+	AssignmentExpression(node, context) {
+		// Track `Component.$$slot = 'slotName'` assignments
+		if (
+			node.left.type === 'MemberExpression' &&
+			node.left.object.type === 'Identifier' &&
+			node.left.property.type === 'Identifier' &&
+			node.left.property.name === '$$slot' &&
+			!node.left.computed &&
+			node.right.type === 'Literal' &&
+			typeof node.right.value === 'string'
+		) {
+			const component_name = node.left.object.name;
+			const slot_name = node.right.value;
+			context.state.slots.set(component_name, slot_name);
+		}
+
+		context.next();
+	},
 };
 
 export function analyze(ast, filename) {
@@ -549,6 +563,7 @@ export function analyze(ast, filename) {
 		ast,
 		scope,
 		scopes,
+		slots: new Map(),
 	};
 
 	walk(
@@ -557,6 +572,7 @@ export function analyze(ast, filename) {
 			scope,
 			scopes,
 			analysis,
+			slots: analysis.slots,
 		},
 		visitors,
 	);
--- a/packages/ripple/src/compiler/phases/3-transform/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/index.js
@@ -628,6 +628,17 @@ const visitors = {
 				if (child.type === 'Component') {
 					const id = child.id;
 					props.push(b.prop('init', id, visit(child, state)));
+				} else if (
+					child.type === 'Element' &&
+					child.id?.type === 'Identifier' &&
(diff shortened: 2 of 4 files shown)


trueadm:
I'm not sold on this approach, it lacks type safety and is another thing to learn, especially when you need to pass props from the child component back out.

What if allowed `component` to have optional params?

```jsx
<Button>
  component A { <div>{"Header"}</div> }
  <div>{"content"}</div>  
  component B { <div>{"Footer"}</div> }
</Button>
```

Now it's far shorter than doing your `$$slot` approach and is much easier to compose.

nihgwu:
@trueadm Yeah I agree this approach is not type safe

Actually I thought about

```jsx
<Button>
  @A { <div>{"Header"}</div> }
  <div>{"content"}</div>  
  @B { <div>{"Footer"}</div> }
</Button>
```

Here `@A` means `at slot 'A'`, and we can even have intellisense when typing `@` in children

`$` -> reactive
`$$` -> slots, so it should be `$$children` instead of `$children`
`@` -> at slot

and it can be shorter with one line element
```jsx
<Button>
  @A <div>{"Header"}</div> 
  <div>{"content"}</div>  
  @B <div>{"Footer"}</div> 
</Button>
```

But the initial problem still exists, from ui lib perspective, we want to hide the details of slots, and replace with sub components

```jsx
<Button>
  <Button.LeadingIcon />
  <Button.Label />
  <Button.TrailingIcon />
</Button>
```

and `Button.*` to encapsulate slot details

nihgwu:
Regarding your proposal, that's counterintuitive as `component` is declaring a component, without `()` it's defining an element, thus we need a new syntax for it

nihgwu:
> Now it's far shorter than doing your $$slot approach and is much easier to compose.

I think you are mislead by the PR's description, I'll update the last example, in practice, the encapsulation is done at ui lib level, not in app

trueadm:
I'm confused with your last example:

```jsx
<Button>
  <Button.LeadingIcon />
  <Button.Label />
  <Button.TrailingIcon />
</Button>
```

You can do this today, they're available as `$children`. You can name them today:

```jsx
<Button>
  component A() { <Button.LeadingIcon /> }
  <Button.Label />
  component B() { <Button.TrailingIcon /> }
</Button>
```

Unless I'm missing something?

Your examples of:

```jsx
<Button>
  @A <div>{"Header"}</div> 
  <div>{"content"}</div>  
  @B <div>{"Footer"}</div> 
</Button>
```

These aren't parser friendly, `@A` would also wrap the subsequent line `<div>{"content"}</div>` unless you added a `;`, because remember these aren't expressions, but statements.

nihgwu:
> You can do this today, they're available as $children. You can name them today:
> 
> ```jsx
> <Button>
>   component A() { <Button.LeadingIcon /> }
>   <Button.Label />
>   component B() { <Button.TrailingIcon /> }
> </Button>
> ```

My answer:
**Imagine you have many buttons on a single page, how it will be**
Full of `component slotName() {}` notations

Regarding `@` syntax, yes you are right, hard to be parsed by acorn, that's just my thought about syntax sugar, it doesn't solve the repetitive inline slot notation issue

trueadm:
We're not going ahead with this. Whilst it might mean less typing in one instance, it also makes for more complication regarding types and understanding of how components as children work. Thanks for the great discussion.

PR #102, opened 2025-09-07
Safe scope

leonidaz:
also fixes `$length` setter on the array to not unnecessarily register deps

### DIFF
--- a/packages/ripple/src/runtime/internal/client/runtime.js
+++ b/packages/ripple/src/runtime/internal/client/runtime.js
@@ -1064,7 +1064,7 @@ export function safe_scope(err = 'Cannot access outside of a component context')
 		throw new Error(err);
 	}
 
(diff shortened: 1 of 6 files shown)


packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -1052,6 +1052,14 @@ export function scope() {
 	return active_scope;
 }
 
+export function safe_scope(err = 'Cannot access outside of a component context') {
+	if (active_scope === null) {
+		throw new Error(err);
+	}
+
+	return scope();
```
trueadm: this can just return `active_scope` right? Also it's missing jsdoc types

leonidaz: done.

PR #107, opened 2025-09-08
Add an edge template and link pkg.pr.new to it

JakeIsMeh:
cc @leonidaz


leonidaz:
I hate to see us duplicating code and then having to update in both, places. could use the github actions to copy the basic template into edge and there should also be a github action to modify the package.json to change the version?

I think we can just a directory `.edge` and add it to the .gitignore at the root of the project.

JakeIsMeh:
Oops, my bad. I checked pkg.pr.new and they seem to do some magic to generate the playground. The edge playground isn't needed.

> With templates, pkg.pr.new will generate Stackblitz instances for the given directories with the new built packages.

templates/basic/package.json
```diff
@@ -1,28 +1,28 @@
...
-	"devDependencies": {
-		"prettier": "^3.6.2",
-		"prettier-plugin-ripple": "0.2.15",
-		"typescript": "^5.9.2",
-		"vite": "^7.1.4",
-		"vite-plugin-ripple": "0.2.15"
-	},
-	"dependencies": {
-		"ripple": "0.2.15"
-	}
```
leonidaz, line 27: @JakeIsMeh could you please revert the white spacing formatting here? should be tabs not spaces.

PR #108, opened 2025-09-08
Auto-publish to OVSX and VSM

JakeIsMeh:
@trueadm some work will be required on your part.

There are two secrets that you'll need to setup, VSM_TOKEN and OVSX_TOKEN;

You'll have to follow the instructions here.
https://github.com/EclipseFdn/open-vsx.org/wiki/Publishing-Extensions

### DIFF
--- a/.github/workflows/vsix.yml
+++ b/.github/workflows/vsix.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [main, develop]
+    branches: [main]
     tags: ["!**"]
     paths: ["packages/ripple-vscode-plugin/**"]
 


trueadm:
I only have one token, the Azure one. Not sure what that comes under either.

JakeIsMeh:
The Azure token is for the Visual Studio Marketplace; goes under VSM_TOKEN.
For Open-VSX, you'll need to acquire a token using the wiki instructions linked, and then set it as OSVX_TOKEN.

trueadm:
You can setup open VSX right, it doesn’t need to be me?

JakeIsMeh:
@trueadm, do you think this should be a manually triggered workflow? I'm thinking of removing the version check.

trueadm:
It can be automated I think. That should be fine

PR #110, opened 2025-09-08
fix: handle single array argument in ripple array constructor

kovsu:
Support `new RippleArray(1, 2, 3)` or `new RippleArray([1, 2, 3])`

### DIFF
--- a/packages/ripple/src/runtime/array.js
+++ b/packages/ripple/src/runtime/array.js
@@ -49,6 +49,11 @@ export class RippleArray extends Array {
 	}
 
 	constructor(...elements) {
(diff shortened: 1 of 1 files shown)


leonidaz:
tbh I’m not sure we should be supporting a non-standard way of constructing an array. The result is unexpected and the philosophy of Ripple is to minimize concepts and learning.  RippleArray.from can already accomplish this and it can also take anything array-like.

I’d say let’s stick to 1:1 with the standard Array.

trueadm:
Agreed. Thanks for the contribution but as mentioned there are better ways here :)

Issue #118, opened 2025-09-10
Variable not defined in parent component, if used with $class attribute


ansarizafar:
The following code works fine, If I use it in a child component but in a parent component, I am getting a strange error.
```jsx
  let $value = 'bg-green-400';
  <p $class={$value}>{'Color value'}</p>
```

<img width="604" height="159" alt="Image" src="https://github.com/user-attachments/assets/17e0cf31-3053-4bfb-b13c-e7c70eb6bac0" />

leonidaz:
code to reproduce, wrapping in a `<div>` causes this error, otherwise it's fine.  Looks like a lexical scope issue.

```jsx
export component App() {
    <div>
        let $value = 'bg-green-400';
        <button onClick={() => $value = generateRandomColor()}>{'Change Color'}</button>
        <p $class={$value}>{'Color value: '}{ $value}</p>
    </div>
}

function generateRandomColor() {
    const randomNumber = Math.floor(Math.random() * 0xFFFFFF);
    let hexColor = randomNumber.toString(16);
    hexColor = hexColor.padStart(6, '0');
    return `#${hexColor}`;
}
```

WebEferen:
I have fixed this issue, the PR is here: https://github.com/trueadm/ripple/pull/122
Seems like there was an issue with referencing the reactive elements on child elements.

ansarizafar:
@WebEferen I can confirm. the issue has been resolved. Thanks

PR #122, opened 2025-09-10
Fix nested variable scoping for reactive attributes

WebEferen:
Modified the Element handler in the transform phase to ensure reactive attribute processing happens within the same scope as child variable declarations:
- Collect reactive attribute updates locally during initial processing
- Defer reactive attribute processing until after children are processed
- Add reactive attributes to child scope so they can access variables declared in children

**Problem**
When declaring reactive variables inside JSX elements and referencing them in reactive attributes on child elements, the compiler would generate JavaScript with incorrect scoping. Variables declared inside JSX elements were not accessible to reactive attributes on child elements within the same scope, causing `ReferenceError: [variable] is not defined` at runtime.
See issue: #118

The compiler was processing reactive attributes ($class, $value, $checked, etc.) outside the child scope block where variables were declared. This created a JavaScript scoping issue where:
- Variable declarations were wrapped in child scope blocks
- Reactive attribute processing happened in the parent scope
- Variables were not accessible across scope boundaries

### DIFF
--- a/packages/ripple/tests/basic.test.ripple
+++ b/packages/ripple/tests/basic.test.ripple
@@ -811,4 +811,25 @@ describe('basic', () => {
 		flushSync();
 		expect(greetingP.textContent).toBe('Hello, User!');
 	});
+
+	it('renders with reactive attributes with nested reactive attributes', () => {
+		component App() {
+			let $value = 'parent-class';
+
+			<p class={$value}>{'Colored parent value'}</p>
+
+			<div>
+				let $nested = 'nested-class';
+
+				<p class={$nested}>{'Colored nested value'}</p>
+			</div>
+		}
+
+		render(App);
+
+		const paragraphs = container.querySelectorAll('p');
+
+		expect(paragraphs[0].className).toBe('parent-class');
+		expect(paragraphs[1].className).toBe('nested-class');
+	});
 });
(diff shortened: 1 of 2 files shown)


packages/ripple/tests/basic.test.ripple
```diff
@@ -811,4 +811,25 @@ describe('basic', () => {
...
+	it('renders with reactive attributes with nested reactive attributes', () => {
+		component App() {
+			let $value = 'parent-class';
+
+			<p class={$value}>{'Colored parent value'}</p>
+
+			<div>
+				let $nested = 'nested-class';
+
+				<p class={$nested}>{'Colored nested value'}</p>
```
leonidaz:
```suggestion
				<p $class={$nested}>{'Colored nested value'}</p>
```

packages/ripple/tests/basic.test.ripple
```diff
@@ -811,4 +811,25 @@ describe('basic', () => {
 		flushSync();
 		expect(greetingP.textContent).toBe('Hello, User!');
 	});
+
+	it('renders with reactive attributes with nested reactive attributes', () => {
+		component App() {
+			let $value = 'parent-class';
+
+			<p class={$value}>{'Colored parent value'}</p>
```
leonidaz:
```suggestion
			<p $class={$value}>{'Colored parent value'}</p>
```

WebEferen: True! Good catch, by the time of pushing it I just accidentally format it. Readded

packages/ripple/tests/basic.test.ripple
WebEferen: Readded - thanks for pointing out as formatted hijacked the $ (need to update my prettier config).

trueadm (approved):
Thank you for this!

Issue #149, opened 2025-09-11
[Prettier] Missing colon before TypeScript return types in methods/functions


e-daigle:
When printing TypeScript return types for functions/methods, the printer omits the `: ` before the return type.

## Minimal example
```ripple
export component FooBar() {
	function Foo(): string {
		return '';
	}
}

class Foo {
	bar(): number {
		return 1;
	}
}
```

## Actual output
```ripple
export component FooBar() {
	function Foo()string {
		return '';
	}
}

class Foo {
	bar()number {
		return 1;
	}
}
```

## Expected output (no change)
```ripple
export component FooBar() {
	function Foo(): string {
		return '';
	}
}

class Foo {
	bar(): number {
		return 1;
	}
}
```

## Proposed fix
I'll try and provide a PR for this. I'll add the colon before both functions and methods return type.

e-daigle:
Please view PR #150

Found a few other minor issues with the prettier plugin, I'll try and document them soon. I know the team is busy working on ripple itself so I'll try and fix as much of the plugin as I can.

WebEferen:
Merged, closing the issue.

PR #150, opened 2025-09-11
fix: add missing colon before method return types (prettier)

e-daigle:
## Summary of changes
- Added `: ` to printFunctionDeclaration/printFunctionExpression/printMethodDefinition when the node contains a return type.

## Testing performed
- All existing tests passed. 
- Tested the example in the issue.
- Tested in a little more complex file; didn't impact the rest of the formatting.

## Any breaking changes
As far as I know, not at all. 😄 

## Related issue numbers
- Issue #149


trueadm:
Thank you for this, please could you also add a test to the Prettier index.test.js module too?

Issue #147, opened 2025-09-11
Mismatched closing tag like <div></span> is incorrectly allowed - parser should raise error


ywnaing:
The parser currently allows mismatched HTML/component tag pairs like:
```
export default component Example() {
  <div></span>
}
```
This compiles successfully, but it should raise a Mismatched closing tag syntax error  similar to how most JSX/HTML parsers behave.

**Expected behavior**

It should fail to compile and return a helpful error message like:

`Mismatched closing tag: expected </div> but got </span>
`

**Actual behavior**

> No error is thrown. The component compiles successfully, and no validation happens for the closing tag name.

**Proposed fix**

I’ve already explored the internals of the RipplePlugin() parser and I’m happy to send a PR for this. My plan is to:

- Add a tagNameToString() helper to stringify tag names from the AST
- Update parseTemplateBody() to compare open/close tag names via this.#path.at(-1)
- Throw a targeted syntax error if they don’t match

Would love any feedback from the maintainers before submitting a patch.

WebEferen:
Thanks for reporting this issue! I have opened a PR that solves it: https://github.com/trueadm/ripple/pull/152

PR #152, opened 2025-09-11
Fix JSX parser to validate matching opening/closing tags

WebEferen:
### Overview

This PR fixes a critical bug in the Ripple compiler's JSX parser where unclosed and mismatched tags were silently accepted instead of throwing compilation errors.

### Problem 
Malformed JSX like `<div></span>` would compile successfully, potentially causing runtime issues

### Solution
- Added tag validation in `parseTemplateBody()` to ensure closing tags match their corresponding opening tags
- Added validation to detect completely unclosed tags when parsing elements
Implemented `getElementName()` helper to extract tag names from JSX nodes for comparison
- Added comprehensive test cases for both mismatched tags (`<div></span>`) and unclosed tags (`<div>content`)

The compiler now throws descriptive errors:
- Expected closing tag to match opening tag. Expected `</div>` but found `</span>`
- Unclosed tag `<div>`. Expected `</div>` before end of component.

Referencing the issue #147

### DIFF
--- a/packages/ripple/tests/basic.test.ripple
+++ b/packages/ripple/tests/basic.test.ripple
@@ -852,7 +852,6 @@ describe('basic', () => {
 			compile(malformedCode, 'test.ripple', 'test.ripple');
 		}).toThrow('Unclosed tag');
 	});
-});
 
 	it('basic reactivity with standard arrays should work', () => {
 		let logs = [];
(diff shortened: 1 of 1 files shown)


WebEferen:
Looking at the tests, it was post merge so need to update them.
@trueadm feel free to check it out once again (post merge issues with missing `}`)

PR #154, opened 2025-09-11
feat: create-ripple-app package for the basic app scaffolding

WebEferen:
This PR introduces a brand new `create-ripple-app` CLI package that provides an interactive setup experience for creating Ripple applications, similar to `create-next-app`, `create-react-app`, and other popular scaffolding tools.

I have registered the package `create-ripple-app` under my name to make sure no one will take it away, so when the time come, feel free to reach out to me and I will just transfer it to You @trueadm.

Package has their own tests and it uses the templates from the main project directory (for now it only has the "basic" one).
Feel free to check it out.

### DIFF
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -88,6 +88,7 @@ jobs:
             './packages/ripple' \
             './packages/rollup-plugin-ripple' \
             './packages/vite-plugin-ripple' \
+            './packages/create-ripple' \
             --compact \
             --comment=update \
             --packageManager=pnpm \


JakeIsMeh:
Thank you for this! Personally, I think it's a bit early for this, and I'd prefer using a dependency tree closer to the vite/vue/svelte CLIs, but I think that's for the team to decide.

leonidaz:
I'm good with this as long as we have enough people to maintain it.

WebEferen:
@leonidaz We anyway have the templates so this is just a small addition that help getting more traction from other people. As it does not have many dependencies in here, it will be more stable - also any change in templates in root directory will be immediately reflected in this package.

It's good to keep the same standards and quick adoption for early users - with that, initial documentation might be shortened to just "use this command and get started".

leonidaz:
i already thought about it and was convinced it's a good idea.  Was just hinting and hoping you'd be around if something goes wrong, lol.

I'd want to @trueadm to sign off as well.

WebEferen:
If there will be any need for additional maintainers, then feel free to reach out to me - I could support it as well :)

leonidaz:
what's your discord handle?

trueadm:
Sounds good. Should we move it to the ripple repro and maintain it? Or keep it external?

WebEferen:
I think it should be kept under ripple repository as templates lives there. For the npm package deployment - I can hand it over to you as I reserved the package name and we can build the pipelines for automatic deployment and release management.

trueadm:
> I think it should be kept under ripple repository as templates lives there. For the npm package deployment - I can hand it over to you as I reserved the package name and we can build the pipelines for automatic deployment and release management.

Sounds good, it's `trueadm` on npm too :)

JakeIsMeh:
while you're at it, do reserve the create-ripple package name too.

WebEferen:
Changed package name to `create-ripple` as it better reflects it's purpose. Thanks @JakeIsMeh 
I will keep `create-ripple-app` name reserved as well.

JakeIsMeh:
@WebEferen , can you also update the pkg.pr.new workflow to publish create-ripple?

PR #155, opened 2025-09-11
trigger playground preview after publishing to pkg.pr.new

hatemhosny:
As [suggested](https://github.com/trueadm/ripple/pull/116#issuecomment-3281539167), this PR changes the trigger of the preview playground in PR to run after publishing to pkg.pr.new (which currently only runs after approval)


leonidaz:
nice! thank you!

PR #159, opened 2025-09-12
add manual trigger for publishing vscode ext

JakeIsMeh:
since the extension is decoupled from the compiler, it sees little churn. This should let us publish it first.

### DIFF
--- a/.github/workflows/vsix-manual.yml
+++ b/.github/workflows/vsix-manual.yml
@@ -1,0 +1,46 @@
+on: workflow_dispatch
+
+name: Publish VSC Extension (Manual)
+jobs:
+    publish:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+
+            - name: Install pnpm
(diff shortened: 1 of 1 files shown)


.github/workflows/vsix-manual.yml
```diff
@@ -0,0 +1,46 @@
...
+
+            - name: Setup Node.js
+              uses: actions/setup-node@v4
+              with:
+                  node-version: 24
+                  cache: "pnpm"
+
+            - name: Install deps w/ npm
+              working-directory: ./packages/ripple-vscode-plugin
+              run: npm ci
```
WebEferen, line 24: Is there any reason why this command uses npm ci instead of pnpm install (as we have cache in place)?

JakeIsMeh: precaution, mostly. it's the only package with a package-lock.json and yarn.lock. most vscode extension guides also only ever show usage of these two.

WebEferen: Sure, just wanted to understand the reasoning

Issue #18, opened 2025-08-31
vscode angry squiggles - block and $ issues


leonidaz:
1. the first error is `Cannot find name '$'. Do you need to install type definitions for jQuery? Try `npm i --save-dev @types/jquery`.ts(2581)`

2. and the second: `Cannot find name '__block'.ts(2304)`

<img width="628" height="260" alt="Image" src="https://github.com/user-attachments/assets/851a513c-7509-4843-8682-5d5cee8940d5" />

code: 

```jsx
export component App() {
  let obj = { $count: 0, $value: 0 };
  let $value = 0;
}

```

_NOTE: Both can happen even within blocks, e.g. `<div>...</div>`_

WebEferen:
Should be fixed in this PR: #165

leonidaz:
@WebEferen as per Jake, there are still a couple of issues remain, as per #165 comments after the merge.  If you have time would you mind looking into them?

JakeIsMeh:
syntax and LSP has gone through a lot of churn, outdated.

PR #165, opened 2025-09-12
Fix ripple vscode plugin block (dollar) highlight

WebEferen:
### Problem
Ripple files using reactive variables with $ prefix (e.g., `let $value = 0, { $count: 0 })` were causing TypeScript linting errors (see issue #18):
- "Cannot find name '$'" when using $ prefixed properties/variables
- "Cannot find name 'block'" for compiler-injected runtime symbols

### Solution
- Enhanced TypeScript definitions in `packages/ripple/types/index.d.ts`:
  - Added global declarations for compiler-injected symbols (__block)
  - Added type definitions for the Ripple runtime namespace ($)
- Fixed VS Code language server compilation in `packages/ripple/src/compiler/index.js`:
  - Enhanced `compile_to_volar_mappings` to include proper imports and type references
  - Added automatic import generation for `ripple/internal/client` when needed
- Improved TypeScript transformation in `packages/ripple/src/compiler/phases/3-transform/index.js`:
  - Fixed variable declaration transformation to work in TypeScript mode `(to_ts: true)`
  - Added proper handling for both simple and complex tracked variable patterns
  - Ensured reactive variables are transformed to `$.tracked()` calls for type safety

Bonus:
- Changed the name of the create-ripple package in README

### DIFF
--- a/packages/create-ripple/README.md
+++ b/packages/create-ripple/README.md
@@ -1,4 +1,4 @@
-# create-ripple-app
+# create-ripple
 
 Interactive CLI tool for creating new Ripple applications.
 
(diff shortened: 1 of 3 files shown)


JakeIsMeh:
just tested this, the dollar signs are indeed gone but it introduces sourcemap drift from the very first new line it encounters.

WebEferen:
@JakeIsMeh it is now fixed, I've redesigned the method and now we will not be dependent on the first sourcemap line that pointed out to the ripple source (now it's purely typescript linter work).

JakeIsMeh:
slight regression: seems to break highlighting of reactive strings
image

JakeIsMeh:
image

it's not angry anymore, but seems like the dollar sign still haunts us

Issue #168, opened 2025-09-12
[Bug]: Updating an array item with a for of loop errors at destroy_block


leonidaz:
### Issue Category

Behavioral Bug

### Bug Description

Error stack:
```
Uncaught TypeError: Cannot read properties of null (reading 'start')
    at destroy_block (chunk-GQ3CYRLU.js?v=3790c65b:1187:25)
    at reconcile (chunk-GQ3CYRLU.js?v=3790c65b:1878:13)
    at chunk-GQ3CYRLU.js?v=3790c65b:1764:19
    at untrack (chunk-GQ3CYRLU.js?v=3790c65b:702:12)
    at Object.fn (chunk-GQ3CYRLU.js?v=3790c65b:1764:5)
    at run_block (chunk-GQ3CYRLU.js?v=3790c65b:322:22)
    at flush_updates (chunk-GQ3CYRLU.js?v=3790c65b:549:13)
    at flush_queued_root_blocks (chunk-GQ3CYRLU.js?v=3790c65b:586:5)
    at flush_microtasks (chunk-GQ3CYRLU.js?v=3790c65b:603:3)
```

### Reproduction Steps

Steps to Reproduce.

1. Click on `change first item`
2. Check the dev tools console, it should have the error

StackBlitz reproduction: https://stackblitz.com/edit/cl9sp7fj?file=src%2FApp.ripple

Code:

```jsx
import { RippleArray } from 'ripple';

export component App() {
  <ArrayTest />
}

component ArrayTest() {
  let items = new RippleArray(1, 2, 3);
  let $sum = 0;

  for (const item of items) {
    $sum += item;
  }

  <button onClick={() => items.push(4)}>{'add item'}</button>
  <button onClick={() => items[0] = 10}>{'change first item'}</button>
  <pre>{JSON.stringify(items)}</pre>
  <pre>{$sum}</pre>
}
```

### Environment

```markdown
- latest ripple version, but present in earlier ones as well
```

### Checklist

- [x] I've searched for existing issues
- [ ] I'm willing to help implement a fix

DipamJha:
import { RippleArray } from 'ripple';

export component App() {
  <ArrayTest />
}

component ArrayTest() {
  let items = new RippleArray(1, 2, 3);
  let $sum = 0;

  for (const item of items) {
    $sum += item;
  }

  <button onClick={() => items.push(4)}>{'add item'}</button>
  
  // Corrected line: Use splice() to ensure reactivity
[...]

PR #171, opened 2025-09-12
Reworking RippleArray to be fully reactive

leonidaz:
- reworks RippleArray to make fully reactive 
- adds full reactive support for missing standard Array properties and methods
- adds full support for reactivity for sparse / holey arrays.
- adds an internal [ARRAY_SET_INDEX_AT] method to emulate setting array via `[]`, needed for `set_property` instead of using `splice()` (it's faster vs adding and then removing via splice)
- add types to RippleArray to make ts happy
- adds a slew of tests for all RippleArray methods and their reactivity
- adds a new static `fromAsync` method to create an array to match the native Array
- ensures setting array's `length` throws an error via the runtime (can't override length in the class)
- fixes a bug where if an array references a non-existent index, once added it wasn't reactive

### DIFF
--- a/packages/ripple/src/runtime/array.js
+++ b/packages/ripple/src/runtime/array.js
@@ -417,11 +417,11 @@ export class RippleArray extends Array {
 			throw new TypeError('Provided index must be a valid integer');
 		}
 
-		index = index < 0 ? index + length : index;
-
-		if (init_index < 0 && index >= length) {
+		if (init_index < -length) {
 			throw new RangeError('Provided negative index out of bounds');
 		}
+
+		index = index < 0 ? index + length : index;
 
 		super[index] = value;
 
--- a/packages/ripple/tests/array.test.ripple
+++ b/packages/ripple/tests/array.test.ripple
@@ -1098,7 +1098,7 @@ describe('RippleArray', () => {
 		expect(container.querySelectorAll('pre')[3].textContent).toBe('40');
 	});
 
-	it.skip('handles ARRAY_SET_INDEX_AT method with reactivity', () => {
+	it('handles ARRAY_SET_INDEX_AT method with reactivity', () => {
 		component ArrayTest() {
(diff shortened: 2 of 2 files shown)


leonidaz:
I should be done at this point.  Need to definitely fix one bug and the other one would be great as well so we can use the for loop in a test.

#175 is a must
#168 would be great use `for of` for testing array updates.

I fixed all other stuff locally so all tests except the above pass.

leonidaz:
ok, at this point, once #175 is fixed this would be good to review and merge.

packages/ripple/src/compiler/phases/3-transform/index.js
```diff
@@ -990,7 +990,8 @@ const visitors = {
 		const left = node.left;
 
 		if (left.type === 'MemberExpression') {
-			if (left.property.type === 'Identifier' && is_tracked_name(left.property.name)) {
+			// need to capture setting length of array to throw a runtime error
+			if (left.property.type === 'Identifier' && (is_tracked_name(left.property.name) || left.property.name === 'length')) {
```
leonidaz, line 994: I hard-coded the `length` property here so we can wrap any `array.length  = 100`, in order to throw an error if length is attempted to be set.  The error on the `set length()` of the array cannot override the native implementation.

Wanted to get your feedback if hard-coding it here is ok.

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -903,6 +903,8 @@ export function get_property(obj, property, chain = false) {
 		var spread_fn = obj[SPREAD_OBJECT];
 		var properties = spread_fn();
 		return get_property(properties, property, chain);
+	} else if (is_ripple_array(obj)) {
+		obj.$length;
```
leonidaz, line 907: fixes a bug where if an array references a non-existent index, once added it wasn't reactive

PR #174, opened 2025-09-12
Add docs

JakeIsMeh:
The other PR seems to have stalled 🥴.

Also using vitepress instead of starlight due to familiarity, lmk if I should switch.

tangent: if there are any weird diffs, lmk. I'm trying out Zed's git integration and it seems somewhat buggy rn,


website/docs/guide/application.md
```diff
@@ -0,0 +1,50 @@
...
+
+```js
+import { mount } from 'ripple';
+// @ts-expect-error: known issue, we're working on it
+import { App } from './App.ripple';
+
+mount(App, {
+	target: document.getElementById('app')!,
+});
+````
```
WebEferen: Remove one tick

WebEferen:
@JakeIsMeh Once you are adding a documentation page, could you also add a custom 404 page? There was an issue opened yesterday about missing, so maybe we could squeeze one in here?  (Ref: #177)

I can open a pr with just this change, but it seems pretty small so maybe there is no point of waiting for the docs and then add the page separately.

JakeIsMeh:
vitepress outputs a 404 page, but if needed I think a custom one can also be added.

WebEferen:
@JakeIsMeh it's about global 404 page (website 404, not the docs as vitepress is actually doing it as you mentioned).
https://www.ripplejs.com/xyz - this one for example.

JakeIsMeh:
the output is flattened, so the 404 page is used for everything. feel free to try it in the preview.

JakeIsMeh:
Sorry folks, it's still sort of a mess, but I haven't had much time to work on it these few days.

JakeIsMeh:
We have a new brand identity, but we don't have the raw files for them yet. Until then, we'll use the current branding.

WebEferen (approved):
Looks good for me. We would need to merge it and then apply the changes on the docs directly - as would be nice to have it deployed already

JakeIsMeh:
Thanks for the review. I'm gonna wait for a second review too since this affects our web presence.

leonidaz (approved):

PR #176, opened 2025-09-12
feat: add highlight for reactive-arrays

WebEferen:
Added highlight option for the reactive arrays.


JakeIsMeh:
I'll review this in a bit, not at my computer rn.

trueadm:
Let's hold of merging this for now.

JakeIsMeh:
confimed to work, but we'll discuss this first

trueadm:
What if we move from `@` to`#`?

WebEferen:
@trueadm that would be great. I can update the syntax to match the # symbol instead. Should I do it?

trueadm:
Go for it, we'll also need to update the parser and compiler eventually too. However, this stuff is low pri right now.

levibassey:
> What if we move from `@` to`#`?

the `#` looks way better

JakeIsMeh:
Superseded by 6834773a739423e0dd0159e3a4688a5e3d8d0f62

PR #184, opened 2025-09-13
Fix up VSC extension workflows

JakeIsMeh:
Successful run:
https://github.com/trueadm/ripple/actions/runs/17694945693


.github/workflows/vsix.yml
```diff
@@ -1,80 +1,106 @@
...
+
+            - name: Build VSIX
+              working-directory: ./packages/ripple-vscode-plugin
+              run: pnpx @vscode/vsce package
+
+            - name: Upload VSIX
+              uses: actions/upload-artifact@v4
+              with:
+                  name: ripple-vscode-plugin VSIX
+                  path: ./packages/ripple-vscode-plugin/*.vsix
```
WebEferen, line 68: Consider using published directory instead of root

.github/workflows/vsix.yml
```diff
@@ -1,80 +1,106 @@
...
+              with:
+                  name: ripple-vscode-plugin VSIX
+
+            - name: Install pnpm
+              uses: pnpm/action-setup@v4
+              with:
+                  version: 10
+
+            - name: Publish to VSM
+              run: pnpx -- @vscode/vsce publish -i *.vsix
```
WebEferen, line 85: If above directory would be changed to published, then this will also need to be updated

.github/workflows/vsix.yml
```diff
@@ -1,80 +1,106 @@
...
+              with:
+                  name: ripple-vscode-plugin VSIX
+
+            - name: Install pnpm
+              uses: pnpm/action-setup@v4
+              with:
+                  version: 10
+
+            - name: Publish to OVSX
+              run: pnpx ovsx publish *.vsix
```
WebEferen, line 104: The same situation here, whether we decide to keep build artifacts outside of root directory, this also need to be changed afterwards.

.github/workflows/vsix.yml
JakeIsMeh: The vsce CLI outputs the built extension to root.
See [here](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#packaging-extensions).

WebEferen: Ohh cool then! Lgtm

Issue #190, opened 2025-09-13
[Bug]: Prettier/formatter is removing commented code on save


abdulrahilsheikh:
### Issue Category

Other

### Bug Description

In a component when we comment a part of code and save the file. The formatter removes the commented code.

1. Code before save

<img width="646" height="556" alt="Image" src="https://github.com/user-attachments/assets/aea24633-b2e1-4ed9-ac8c-e2ef63aa0584" />

2. After save the commented part gets removed

<img width="646" height="556" alt="Image" src="https://github.com/user-attachments/assets/4f84a0e3-02cd-4b17-9793-fbd10ff398fb" />

### Reproduction Steps

1. In any component comment a part of the code.
2.  Save the file.

### Environment

```markdown
- macOS 15.6
- vscode
```

### Checklist

- [x] I've searched for existing issues
- [x] I'm willing to help implement a fix

e-daigle:
Working on a fix for this.

Edit: It was partially implemented in printRippleNode but often skipped because many switch cases return early. Even after refactoring cases to set `nodeContent` rather then returning, placement is still unreliable, especially deep in HTML and even more so in CSS. Could we delegate CSS to Prettier’s parser?

abdulrahilsheikh:
@e-daigle looks like we both worked on same issue🤣. Kinda my bad I didn't mention I'm working on it. Let's merge your PR. I'm searching for more bugs. Let's solve this and move to prod🔥.

e-daigle:
@abdulrahilsheikh I encourage you to share any progress you made. My PR is not complete and there are a lot of bugs still impacting comments.

trueadm:
Okay Prettier comments should be improved.

leonidaz:
@abdulrahilsheikh could you please verify with the latest version if the issue is resolved?

abdulrahilsheikh:
Guys i've checked the new version `"prettier-plugin-ripple": "^0.2.77"`
Its working great the only issue it has when there is a comment before or after any variable declaration or scope.

For ex: before formatting
```
// Some before comments
const test = () => {
	// for (const /* Unknown: ArrayPattern */ of items.entries()) {}
};

// style block gets different spaces/tabs assigned
component AppStyles({ name }) {
	// style block gets different spaces/tabs assigned
	<style>
	</style>
}
```
after formatting
```
const test = () => {
	// for (const /* Unknown: ArrayPattern */ of items.entries()) {}
};

// style block gets different spaces/tabs assigned
[...]

PR #195, opened 2025-09-14
fix: keep comments when formatting (prettier)

e-daigle:
### prettier-plugin-ripple/src/index.js
- Fixed always-truthy condition in `getVisitorKeys`
- Moved the big switch out of `printRippleNode` into `printRippleNodeNoComments`. Simplest way I found to keep the existing comment printing logic. (They were often skipped because many switch cases returned early rather than updating `nodeContent`)
- Fixed indentation in CSS. The second and subsequent rules had 2 extra spaces:
  ``` 
    div {
      color: red;
    }
      div {
      color: red;
    }
  ```
  Fixed:
  ``` 
    div {
      color: red;
    }
    div {
      color: red;
    }
  ```

### prettier-plugin-ripple/src/index.test.js
- Modified `should handle style tags inside component body` to fit the corrected indentation.

### ripple/src/compiler/phases/1-parse/index.js
- Fixed the start index for nodes of type 'Text' that have a comment in front of them. (The nodes had the start index of the comment. When re-adding the comments, they got added to the child 'Literal' node, not the 'Text' node.)

I want to discuss this one more. I explained a little more on discord but I can open an issue. 

## Testing performed
- All existing tests passed. 

## Any breaking changes
None 

## Related issue numbers
- Issue #190


e-daigle:
@trueadm could you take a quick look at 1-parse/index.js? I’m not fully confident in my approach and I'd appreciate your feedback, open to a better fix. Happy to explain what I tried to fix a little more if needed.

packages/prettier-plugin-ripple/src/index.js
```diff
@@ -25,11 +25,11 @@ export const parsers = {
 		},
 
 		locStart(node) {
-			return node.loc.start.index;
+			return node.start;
```
trueadm: Why are these changes needed?

e-daigle: After fixing getVisitorKeys, some tests failed because node.loc.start/end.index was null on specific nodes. I checked Svelte's plugin and saw it uses node.start and that fixed the issue.

trueadm: We might need to update some parts of the parser in this PR then. Specifically around elements. We fix their start position manually but only do that on the loc object. We’d need to apply the same to start too

e-daigle: I guess we could use loc by default but have a condition so that if it's null, we use start.

abdulrahilsheikh:
@e-daigle i took the pull of your code and did some test to it. 
So its producing some inconsistant result.
For below code block when i run format.
 ```
 function A(){
	if ("true")// something to say
	{}
	for (let i = 0; i < 5; i++ ) // something to say
	{}
}
```

The output comes 
```
function A() {
	if ("true" // something to say) {}
	for (let i = 0; i < 5; i++ // something to say) {}
}
```
So i dig a little deeper. So when we run any print statement most of them have like `parts.push(path.call(print, 'test'));` so this cause the `printRippleNode` to run. And at the bottom of this code we have block for getting trailing block.

```
function printIfStatement(node, path, options, print) {
	const parts = [];
	parts.push('if (');
	parts.push(path.call(print, 'test'));
	parts.push(') ');
```
	
But if we log `node.test` in `printIfStatement` it shows that `test` object also have a `trailingComments` field.
this causes that comment `// something to say` getting appended to "true" in if statement.

Similar thing is going with `printForStatement` in this case there is `node.update` which also have `trailingComments` field.
So there is a pattern here.
so whatever is there in parentheses, the last item inside that parentheses have  `trailingComments` field.
Same goes for switch statement 

```	
switch (SomeCondition // some-comment) {
default:
	return;
}
```
It has discriminant in parentheses which again has `trailingComments` in it
That my findings.
So we need some kind of fix for this.
So the solution to this is that we have to go in each printFunction and print the comment before the statements. Also we will need to make the `printRippleNode` skip the trailing comments when it should not run. We can leverage `args` parameter of `printRippleNode`.

**Comment by e-daigle** [comment]
@abdulrahilsheikh thanks for your example and explanation. The parser does add `// something to say` to the Literal `"true"` (the test node of the IfStatement). How will you differentiate those two comments `if ("true" /*1*/) //2` which will both be trailing comments of the "true" node? I can't even find a way inside the parser 


From what I've tested outside the plugin:
```
if ("true")// something to say
{}
```
should become
```    
if ("true") {
      // something to say
}
```

**Comment by abdulrahilsheikh** [comment]
Hi @e-daigle i was looking for fix for `if ("true" /*1*/) //2` as you mentioned above. 
So I checked the official prettier plugin to know how they're handling this edge case. 
https://github.com/prettier/prettier/blob/5067313d6899596c06cb2cad8b19cf87a3febd15/src/language-js/comments/handle-comments.js#L187

So they say that there is no way to handle this using AST or locations of nodes. You can reference the code block in above link. And the official repo of prettier can help us around edge cases. 

As mentioned in above prettier codebase we have to look for next character to se if is a ") "

But in our plugin we are manually printing the end parentheses. So we have to play around a little. I'll try to get a solution for this once I get back to my system.

**Comment by e-daigle** [comment]
@abdulrahilsheikh Here's a dirty quick fix I tried if you want to test: 

In the parser's index.js, around line 740, add a condition to check if there's a ')', kind of like prettier does.
```
else if (node.end <= comments[0].start && /^[,) \t]*$/.test(slice)) {
          if(!(parent && parent.test === node && slice.indexOf(")") > -1) &&
	          !(parent && parent.update === node && slice.indexOf(")") > -1)) {
	          node.trailingComments = [/** @type {CommentWithLocation} */ (comments.shift())];
          }
	}
```

**Comment by abdulrahilsheikh** [comment]
@e-daigle hi i've tested the code. It worked perfectly for  below case.
Before format
```
function A() {
	if ('true' /* asdsd */)// something to say
	{}
	for (let i = 0; i < 5; i++ /* asdasdd*/) // something to say
	{}
	switch (2 /* asdasdd*/) // switch
	 {
		default:
	return;
}
}
```
After Format
```
function A() {
	if ('true' /* asdsd */) // something to say
	{}
	for (let i = 0; i < 5; i++ /* asdasdd*/) // something to say
	{}
	switch (2 /* asdasdd*/) {
	// switch
		default:
	return;
}
}
```

**Comment by e-daigle** [comment]
Great, I wish we could get it to do 
```
if ('true' /* asdsd */) {
  // something to say
}
```
but the result you got is still acceptable.

I'm going to start working on comments in CSS.

e-daigle:
@trueadm I have some changes I'd like you to look at if you have time. I could add them to this PR if needed but I'm probably going to scrap this PR and start a new one after you guys change the reactive language. If you think the issues I are worthy of a fix, you could fix them in your way while working on the compiler (or use some of the changes I provided), I could then fix the prettier issues.

[https://github.com/e-daigle/ripple/pull/1/files](https://github.com/e-daigle/ripple/pull/1/files)

I'd like you to look mainly at 1-parse/index.js and 1-parse/style.js. 

index.js:
- Only one trailing comment could be handled per node. Now it can handle multiple.
- Trailing comments were problematic for `if` and  `for` statements, see the above examples if needed. Now comments after the closing parenthesis become leading comments of the next node.
- Passing base index and `onComment` handler to the style parser for the reasons below

in style.js:
- Comments were simply removed and couldn't be added later on. Now passing `onComment` handler to store the comments.
- Each css node was added to the AST with the wrong start and end values. It was not using the absolute index, rather a relative index from the start of the stylesheet. Now passing a base index to the style parser to

trueadm:
Those changes look good. Scrap this PR and go with that approach if that's what you want, it makes sense too!

trueadm:
@e-daigle When are you going to put up the new PR? I eager to see the changes :)

PR #197, opened 2025-09-14
fix source location for style tag

### LINKED ISSUE #103: Sourcemaps returned from compiler is inaccurate after style tag
Original Title (kept for search): Syntax hightlighting breaks down after style tag and other random ways

Repro is kind of weird, if I reduced it any further, the breakage isn't as viral.
It also extends past the ending brace.

Also, the style tag has to have at least 2 lines within it. If there's only one line, it doesn't break.

```ripple
export component Editor() {
	<style>
		.myStyle {}

	</style>


	ReallyLongFunctionNamePleaseWork()

	const help = myClass.test();

	function handler() {
		
	}

	let myVar = DoSomeWork();

	/** ... Rest of code */
}

function myFunc() {
	doMoreWork();
}
```

Gabriel3232:
Fixes #103 

Problem: The css content was skipped when computing source locations, leading to wrong values starting at the closing `style` tag.

Solution: Adjust the curLine and lineStart manually to take new lines into account.

I'm open to adding some tests in compiler.test.ripple if this approach seems ok

### DIFF
--- a/packages/ripple/src/compiler/phases/1-parse/index.js
+++ b/packages/ripple/src/compiler/phases/1-parse/index.js
@@ -2,6 +2,7 @@ import * as acorn from 'acorn';
 import { tsPlugin } from 'acorn-typescript';
 import { parse_style } from './style.js';
 import { walk } from 'zimmerframe';
+import { regex_newline_characters } from '../../../utils/patterns.js';
 
 const parser = acorn.Parser.extend(tsPlugin({ allowSatisfies: true }), RipplePlugin());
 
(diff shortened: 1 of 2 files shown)


JakeIsMeh (dismissed):

JakeIsMeh:
lovely, fix works like a charm; we'll wait on feedback from the rest of the team.

WebEferen:
Looks good to me, I'd also add tests (at least one test case)

trueadm (approved):
Thank you!

PR #200, opened 2025-09-14
fix: app prefix removal from create-ripple package

WebEferen:
- Removed unused create-ripple-app (app suffix).
- Bumped package version.
- Added missing newlines


templates/basic/.gitignore
JakeIsMeh, line 1: unrelated change

templates/basic/ripple.d.ts
JakeIsMeh, line 1: unrelated change

templates/basic/.gitignore
WebEferen: I just added a newline there (there was missing newline).

templates/basic/ripple.d.ts
WebEferen: Added newline (missing newline within the file).

WebEferen: Can add it in separate PR, but I think it is pointless to create a separate PR just for this.

PR #210, opened 2025-09-15
fixes bracketed function calls #175

### LINKED ISSUE #175: [Bug]: Calling a function that is an object property notation loses this context
### Issue Category

Compilation Bug

### Bug Description

When calling an object, such `obj[myProp]()`, the `this` context is lost.  It seems that the `$.get_property` call returns the function itself and then calls it.  So, the object instance context is lost.

So instead of `obj[myProp]()`, it does: `var fn = obj[myProp]; fn()`

and get compiled to:

```js
var on_click = (_, obj, __block) => $.with_scope(__block, () => $.get_property(obj, 'test')());
```

So, the `$.get_property()` wrapper should not apply for function calls via object props, and should just be:

```js
var on_click = (_, obj, __block) => $.with_scope(__block, () => obj['test']());
```

The minimal repro is below.  For simplicity I replaced the Symbol property that I was using with a regular one.

```jsx
	export component App() {
		const obj = {
			val: 2,
			test: function() {
				return this.val * 2;
			},
		};

		<button onClick={() => obj['test']()}>{'Double Value'}</button>
		<pre>{obj.val}</pre>

	}
```

### Reproduction Steps

Minimal reproducible code is above.

If you run it and click the button, error occurs as `this` is undefined in the console of the dev tools.

```
events.js:142 Uncaught TypeError: Cannot read properties of undefined (reading 'val')
    at test (App.ripple?import&t=1757712774246:24:16)
    at App.ripple?import&t=1757712774246:8:92
    at Module.with_scope (runtime.js:1125:10)
    at HTMLButtonElement.on_click (App.ripple?import&t=1757712774246:8:39)
    at HTMLDivElement.handle_event_propagation (events.js:117:10)
```

### Environment

```markdown
- latest ripple version, probably in earlier ones too
```

### Additional Context

_No response_

### Checklist

leonidaz:
fixes #175 

- introduces `call_property` for handling bracketed function calls

**Better viewed with hide whitespace**

- currently blocks #171

### DIFF
--- a/packages/ripple/src/compiler/phases/3-transform/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/index.js
@@ -150,40 +150,57 @@ const visitors = {
 		}
 
 		// Handle array methods that access the array
-		if (
-			callee.type === 'MemberExpression' &&
-			!callee.optional &&
-			callee.property.type === 'Identifier'
-		) {
-			const name = callee.property.name;
-			if (
-				// TODO support the missing array methods
-				name === 'reduce' ||
-				name === 'map' ||
-				name === 'forEach' ||
-				name === 'join' ||
-				name === 'includes' ||
-				name === 'indexOf' ||
-				name === 'lastIndexOf' ||
-				name === 'filter' ||
-				name === 'every' ||
-				name === 'some' ||
-				name === 'toSpliced' ||
-				name === 'toSorted' ||
-				name === 'toString' ||
-				name === 'values' ||
-				name === 'entries'
-			) {
+		if (callee.type === 'MemberExpression') {
+			const property = callee.property;
+
+			if (property.type === 'Identifier' && !callee.optional) {
+				const name = property.name;
+				if (
+					// TODO support the missing array methods
+					name === 'reduce' ||
+					name === 'map' ||
+					name === 'forEach' ||
+					name === 'join' ||
+					name === 'includes' ||
+					name === 'indexOf' ||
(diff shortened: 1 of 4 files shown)


packages/ripple/src/compiler/phases/3-transform/index.js
```diff
@@ -150,40 +150,57 @@ const visitors = {
...
+								'$.array_' + name,
+								context.visit(callee.object),
+								...node.arguments.map((arg) => context.visit(arg)),
+							),
+						),
+					);
+				}
+			}
+
+			if (callee.computed) {
```
leonidaz, line 190: only implemented this for computed / bracketed function calls, the code below will handle non-computed ones.

packages/ripple/src/compiler/phases/3-transform/index.js
```diff
@@ -150,40 +150,57 @@ const visitors = {
...
+						b.thunk(
+							b.call(
+								'$.array_' + name,
+								context.visit(callee.object),
+								...node.arguments.map((arg) => context.visit(arg)),
+							),
+						),
+					);
+				}
+			}
```
leonidaz, line 188: we should probably also do the same with tuple arrays (just wrap in `call_property`) and handle this in runtime.  I was thinking to attach another unique symbol or just one that's unique on tuples to identify them in runtime and there just apply whatever methods are being called on them.

But better to wait to decide what we'll do with our recent proposals.

PR #217, opened 2025-09-15
fix: apply styles based on reactive attributes

### LINKED ISSUE #194: [Bug]: reactive $class changes are not being applied
### Issue Category

Behavioral Bug

### Bug Description

Only one line become red if the `class` is used vs `$class`.  If `$class` is used then both lines are not red

```jsx
component ArrayTest() {
	let $a = 'red';
	let $b = '';

	<div class={$a} onClick={() => $a ==='red' ? $a = '' : $a = 'red'}>{'works'}</div>
	<div $class={$b} onClick={() => $b ==='red' ? $b = '' : $b = 'red'}>
		{'click to make red: '}
		{$b || 'empty class'}
	</div>

	<style>
		.red { background: red }
	</style>
}
```

and here both don't work as they both use `$class`

```jsx
component ArrayTest() {
	let $a = 'red';
	let $b = '';

	<div $class={$a} onClick={() => $a ==='red' ? $a = '' : $a = 'red'}>{'works'}</div>
	<div $class={$b} onClick={() => $b ==='red' ? $b = '' : $b = 'red'}>
		{'click to make red: '}
		{$b || 'empty class'}
	</div>

	<style>
		.red { background: red }
	</style>
}
```

reported in comment on #5 

### Reproduction Steps

Click on each line to toggle red class.

### Environment

```markdown
- latest and likely earlier
```

### Additional Context

_No response_

### Checklist

moka-ayumu:
Fixes #194 

This PR fixes a bug where the compiler did not correctly handle attributes starting with a dollar sign, such as `$class`.
`attribute_matches` function has been updated to correctly recognize attributes prefixed with a dollar sign.

### DIFF
--- a/packages/ripple/src/compiler/phases/2-analyze/prune.js
+++ b/packages/ripple/src/compiler/phases/2-analyze/prune.js
@@ -405,7 +405,9 @@ function attribute_matches(node, name, expected_value, operator, case_insensitiv
 		if (attribute.type === 'SpreadAttribute') return true;
 
 		if (attribute.type !== 'Attribute') continue;
-		if (attribute.name.name.toLowerCase() !== name.toLowerCase()) continue;
+
+		const lowerCaseName = name.toLowerCase();
+		if (![lowerCaseName, `$${lowerCaseName}`].includes(attribute.name.name.toLowerCase())) continue;
 
 		if (expected_value === null) return true;
 
--- a/packages/ripple/tests/basic.test.ripple
+++ b/packages/ripple/tests/basic.test.ripple
@@ -76,6 +76,12 @@ describe('basic', () => {
 
 			<button onClick={() => $active = !$active}>{'Toggle'}</button>
 			<div $class={$active ? 'active' : 'inactive'}>{'Dynamic Class'}</div>
+
+			<style>
+				.active {
+					color: green;
+				}
+			</style>
 		}
 
 		render(Basic);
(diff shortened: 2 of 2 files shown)


leonidaz:
Could you add a test please?

moka-ayumu:
> Could you add a test please?

@leonidaz 
I've updated the test code to check for classes starting with `ripple-` because the style tag isn't applied in the test environment, which prevented using `getComputedStyle` to check for applied styles.

packages/ripple/src/compiler/phases/2-analyze/prune.js
```diff
@@ -405,12 +405,15 @@ function attribute_matches(node, name, expected_value, operator, case_insensitiv
...
+
+		const lowerCaseName = name.toLowerCase();
+		if (![lowerCaseName, `$${lowerCaseName}`].includes(attribute.name.name.toLowerCase())) continue;
 
 		if (expected_value === null) return true;
 
 		if (is_text_attribute(attribute)) {
-			return test_attribute(operator, expected_value, case_insensitive, attribute.value.value);
+			if (test_attribute(operator, expected_value, case_insensitive, attribute.value.value))
+				return true;
```
leonidaz: this should be reverted

leonidaz (changes requested):
ok, just a small change and then we can merge.

leonidaz (approved):

leonidaz:
thank you!  

Note: in the future after a review, please add another commit vs just pushing in the same ones as it no longer possible to verify the requested changes.

We typically (100% at this point) squash commits anyway.

Issue #213, opened 2025-09-15
[Bug]: Backtick string delimiters without any variables inside cause $children not to render


chandani-rathi:
### Issue Category

Compilation Bug

### Bug Description

on custom component, I am trying to render the formatted string, but hardcoded values are not working.
```
export component Heading({ $level, $children }) {
    if($level == 1) {
        <h1><$children /></h1>
    }
    if($level == 2) {
        <h2><$children /></h2>
    }
    if($level == 3) {
        <h3><$children /></h3>
    }
}
```
below code is not working, I am seeing empty tag in html and values are set in comments tag
```
export default component App() {
  <Heading $level={1}>{`Heading 1`} </Heading>
}

```

Below cases are working
```
export default component App() {
  <Heading $level={1}>{'Heading 1'} </Heading>
}
```
```
export default component App() {
  <Heading $level={1}>{"Heading 1"} </Heading>
}
```
```
export default component App() {
  const $level = 1
  <Heading $level={$level}>{`Heading ${$level}`} </Heading>
}
```

### Reproduction Steps

1. Create custom component to show a string.
2. pass the string using string template.

### Environment

```markdown
- latest
```

### Checklist

- [x] I've searched for existing issues
- [ ] I'm willing to help implement a fix

leonidaz:
looks like the issue is with backticks without any var placeholders.  If it's just a pure string, the $children is not rendered.  With var placeholders, it works.

WebEferen:
Fixed on the PR #218

WebEferen:
Closing issue as change was merged onto main.

leonidaz:
or at least it will be shortly :)

leonidaz:
Reopening as the issue reportedly in #221 still exists

leonidaz:
@WebEferen still seems to be an issue reported again in #221 but i reopened this one and close the new one as a duplicate.

WebEferen:
Great, on it. Thanks @leonidaz, this seems like issue with compiler rather than template literal.

WebEferen:
@leonidaz prepared a fix together with this particular test case. Seems like they are not fixed.

PR #218, opened 2025-09-15
fix: empty template literal bug

WebEferen:
Fixes empty literal bug (backticks).
For reference see issue: #213


leonidaz (approved):
nice!

PR #220, opened 2025-09-15
change decorator api to refs

trueadm:
The `{@use}` decorator feature has always felt cumbersome, as people associate capturing the DOM element using a `ref`. Plus the `@` isn't something we've promoted before, so it doesn't make sense to use it here either. We can simplify this this feature to be `{ref fn}` or `{ref () => {…} }` or `{ref fn(a, b}`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,6 +649,41 @@ Lastly, you can use refs on composite components.
 
 When passing refs to composite components (rather than HTML elements) as shown above, they will be passed a `Symbol` property, as they are not named. This still means that it can be spread to HTML template elements later on and still work.
 
+#### createRefKey
+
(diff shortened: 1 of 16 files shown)


sreekar339339:
would the typical attribute assignment syntax `ref={fn}` not work? `<input ref={fn} />`

JakeIsMeh (approved):

trueadm:
> would the typical attribute assignment syntax `ref={fn}` not work? `<input ref={fn} />`

No. We explicitly don’t want that API as it’s far too limiting - you can only have one ref unless you get into the business of wrapping refs, which is a source of memory leaks and debug hell

PR #230, opened 2025-09-16
ci: add npm provenance check for dependencies

JakeIsMeh:
While we've avoided the recent supply chain attacks through pnpm's default of having to manually approve postinstalls, I think it's prudent that we take an active approach to supply-chain security.


trueadm (approved):
good thinking

PR #239, opened 2025-09-18
Update basic test cases to match proposal

WebEferen:
Updates basic test files to cover new `tracked()` template (tracked for now - it is subject to change).


packages/ripple/tests/accessors-props.test.ripple
```diff
@@ -66,70 +62,17 @@ describe('prop accessors', () => {
     expect(logs).toEqual(['App effect', 0, 'App effect', 1, 'App effect', 2]);
   });
 
-  it('should render a basic prop accessor on a composite component #2', () => {
```
leonidaz, line 69: we should skip this on for now since I don't think reactivity works with composite components yet instead of deleting?

WebEferen: Actually this is the case where the bottom-up communication was set to null (so no handler). With the composites now, once you pass a tracked value and update it downstream, it will always be reflected on the parent component (this might change when you use untrack on prop probably).

WebEferen: I revoked this test and added note there.

trueadm: We shouldn't be disabling or skipping and tests, let alone changing how they work.

trueadm: In fact this file has already been updated on main

PR #249, opened 2025-09-19
Update basic template with new syntax

### LINKED ISSUE #242: Out of date dependencies in basic template
I was trying out Ripple by creating a new project from the template, as suggested by the "Try Ripple" section in the readme. I quickly noticed that DOM attributes were not reactive in my project, but worked properly in the playground. 

```ripple
  let $state = false;
  <button
    onClick={() => $state = !$state}
    data-state={$state} {/* <-- does not update */}
  >
    {$state ? 'ON' : 'OFF'}
  </button>
}
```

I managed to fix it on my end by running `npm update` and restarting. I would recommend updating the package-lock file in the basic template, so that other people trying it out would have a slightly smoother experience.

JakeIsMeh:
Also, Closes #242 

Updated basic template with new syntax
Removed package-lock.json so that each install can have the latest version.


trueadm:
Already done it seems.

Issue #222, opened 2025-09-16
[Feature]: tick()-like function for awaiting UI update caused by preceding state updates.


sreekar339339:
### Feature Category

Developer Experience

### Problem Statement

As a user, I want to run some code like adding, shifting focus etc after state update statements in the same event handler function.
For this to work, I need a function like `tick()` in Svelte that allows me to await for the UI update caused by the state update statements.

### Proposed Solution

a new API called tick() or something similar that is async which resolves when the UI update caused by preceding state update statements is committed to the DOM.

### Checklist

- [x] I've searched for existing feature requests
- [ ] I'm willing to help implement this feature

JakeIsMeh:
It already exists as `flushSync()`.

trueadm:
That’s not quite the same. flushSync forces updates to sync update, whilst something like tick would wait

JakeIsMeh:
My bad.

leonidaz:
resolved (except SSR) by #254

PR #254, opened 2025-09-20
feat: tick

moka-ayumu:
#222 
Added `tick` function.

```ts
import { track, tick } from 'ripple';

export component App() {
	let value = track(0);
	setTimeout(() => @value++, 1000);
	tick().then(() => console.log('rendered'));

	<p>{@value}</p>
}
```

### DIFF
--- a/packages/ripple/src/runtime/internal/client/runtime.js
+++ b/packages/ripple/src/runtime/internal/client/runtime.js
@@ -563,6 +563,12 @@ function flush_queued_root_blocks(root_blocks) {
   }
 }
 
+/** @type {PromiseWithResolvers<void>} */
+let tick_deferred = Promise.withResolvers();
+export function tick() {
+  return tick_deferred.promise;
+}
+
 function flush_microtasks() {
   is_micro_task_queued = false;
 
@@ -585,6 +591,8 @@ function flush_microtasks() {
     flush_count = 0;
   }
   old_values.clear();
+  tick_deferred.resolve();
+  tick_deferred = Promise.withResolvers();
 }
 
 /**
--- a/packages/ripple/src/runtime/index.js
+++ b/packages/ripple/src/runtime/index.js
@@ -36,7 +36,7 @@ export function mount(component, options) {
 
 export { create_context as createContext } from './internal/client/context.js';
 
-export { flush_sync as flushSync, untrack, deferred } from './internal/client/runtime.js';
+export { flush_sync as flushSync, untrack, deferred, tick } from './internal/client/runtime.js';
 
 /**
  * @param {any} v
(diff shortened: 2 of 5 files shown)


leonidaz:
@trueadm fyi, this is what svelte uses.  Do you want to do something different / better?

```js
/**
 * Returns a promise that resolves once any pending state changes have been applied.
 * @returns {Promise<void>}
 */
export async function tick() {
	if (async_mode_flag) {
		return new Promise((f) => requestAnimationFrame(() => f()));
	}

	await Promise.resolve();

	// By calling flushSync we guarantee that any pending state changes are applied after one tick.
	// TODO look into whether we can make flushing subsequent updates synchronously in the future.
	flushSync();
}
```

trueadm:
We should probably do what Svelte does here as it's technically more correct.

moka-ayumu:
@leonidaz @trueadm 
The `async_mode_flag` part doesn't seem necessary at this time, so I haven't added it. Please check if it's good.

leonidaz:
@trueadm our components do support `async` via `try / pending`, so probably safer if we use the `requestAnimationFrame` version, right?

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -563,6 +563,12 @@ function flush_queued_root_blocks(root_blocks) {
   }
 }
 
+// https://github.com/sveltejs/svelte/blob/ded13b825d7efcdf064fd65a5aa9e7e61293a48b/packages/svelte/src/internal/client/runtime.js#L501
+export async function tick() {
+  await Promise.resolve();
+  flush_sync();
+}
```
leonidaz, line 611:
```suggestion
/**
 * @returns {Promise<void>}
 */
export async function tick() {
	return new Promise((f) => requestAnimationFrame(() => f()));
	flushSync();
}
```

since we have try / pending and allow awaits in components

leonidaz (changes requested):
so neet this for async / await.

also, since you have conflicts please rebase on main and we'll merge.

thank you!

moka-ayumu:
I did not include the `flush_sync` code as it had already returned before this line. Please check.

leonidaz:
Right, yes, flushSync is not needed.

JakeIsMeh:
@moka-ayumu, are you still interested in this PR? If so, please rebase your PR before we can merge.

moka-ayumu:
@JakeIsMeh resolved conflicts.

@leonidaz Is it possible to approve it?

leonidaz:
We also need one for SSR but I’m not sure if it should do anything there in async rendering @trueadm

trueadm (approved):

trueadm:
We can get around to async when the time is right

PR #256, opened 2025-09-20
fix: events in spread props

### LINKED ISSUE #255: [Bug]: Event listeners don't work in spread props
### Issue Category

Compilation Bug

### Bug Description

Spreading events from on object onto a dom element, like `onClick` get spread like string vs event handlers

### Reproduction Steps

https://stackblitz.com/edit/kjqepujt?file=src%2FApp.ripple

### Environment

```markdown

```

### Additional Context

_No response_

### Checklist

anubra266:
Fixes #255

### DIFF
--- a/packages/ripple/src/compiler/phases/2-analyze/index.js
+++ b/packages/ripple/src/compiler/phases/2-analyze/index.js
@@ -4,7 +4,6 @@ import { create_scopes, ScopeRoot } from '../../scope.js';
 import {
   get_delegated_event,
   is_element_dom_element,
-  is_event_attribute,
   is_inside_component,
   is_ripple_import,
   is_void_element,
@@ -13,6 +12,7 @@ import { extract_paths } from '../../../utils/ast.js';
 import is_reference from 'is-reference';
 import { prune_css } from './prune.js';
 import { error } from '../../errors.js';
+import { is_event_attribute } from '../../../utils/events';
 
 function mark_control_flow_has_template(path) {
   for (let i = path.length - 1; i >= 0; i -= 1) {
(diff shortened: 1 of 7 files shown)


leonidaz:
So, we have some helper functions and some missing.  Even have one that we use which doesn't event exist:

This one below we're calling from `get_delegated_event` in `packages/ripple/src/compiler/utils.js`  but it's just missing in the codebase 🫨 So, we should fix this first, I guess we either never hit this path or the error gets eaten.
```js
/**
 * @param {string} event_name
 */
function get_attribute_event_name(event_name) {
	event_name = event_name.slice(2);
	if (is_capture_event(event_name)) {
		event_name = event_name.slice(0, -7);
	}
	return event_name;
}
```

This one I pulled from svelte that is being called by the above.
```js
/**
 * @param {string} name
 */
export function is_capture_event(name) {
	return name.endsWith('Capture') && name.toLowerCase() !== 'gotpointercapture' && namename.toLowerCase() !== 'lostpointercapture';
}
```

We already have this one defined in `packages/ripple/src/compiler/utils.js`
```js
export function is_event_attribute(attr) {
  return attr.startsWith('on') && attr.length > 2 && attr[2] === attr[2].toUpperCase();
}
```

anubra266:
@leonidaz the spread props is run time though, do we wanna use the `is_event_attribute` from the compiler utils in the spread props logic? Or I might be misunderstanding your suggestion

anubra266:
@leonidaz just encountered an issue. Every other event handler I tested works well, but `onFocus` does not when used with spread props. Not sure what is different yet

anubra266:
@leonidaz focus bug was because focus doesn't delegate. Fixed that and cleaned up the events.

trueadm:
Thanks for this, I can fix the last piece :)

PR #257, opened 2025-09-21
proxify all new arrays created via TrackedArray

leonidaz:


packages/ripple/src/runtime/array.js
```diff
@@ -55,26 +55,31 @@ TrackedArray.fromAsync = async function (arrayLike, mapFn, thisArg) {
...
+ * @param {{
+ *  elements: Iterable<T>,
+ *  block: Block,
+ *  from_static?: boolean,
+ *  use_array?: boolean
+ * }} params
  * @returns {TrackedArray<T>}
  */
-function proxy(elements, block, is_from_static = false) {
+function proxy({ elements, block, from_static: from_static = false, use_array = false }) {
```
JakeIsMeh, line 71:
```diff
-- from_static: from_static = false
++ from_static = false
```

leonidaz: damn copilot 😂

PR #259, opened 2025-09-21
bundle cli

JakeIsMeh:
I haven't added a `build` command to the monorepo's package.json, mostly because it's only this package that requires a build step. lmk if I should.

before:
```
ripple/packages/create-ripple $ pnpx npm-size .
  √ current directory 2.43 MB
```
after:
```
ripple/packages/create-ripple $ pnpx npm-size .
  √ current directory 403.44 KB
```


WebEferen (approved):
Looks good! Awesome work

PR #272, opened 2025-09-22
Adds on for direct event handling

leonidaz:
- also updates llms.txt for the `<@component>` transport reactivity

### DIFF
--- a/packages/ripple/src/runtime/internal/client/events.js
+++ b/packages/ripple/src/runtime/internal/client/events.js
@@ -16,6 +16,20 @@ var all_registered_events = new Set();
 var root_event_handles = new Set();
 
 /**
+ * @param {EventTarget} element
+ * @param {string} type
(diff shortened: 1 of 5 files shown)


packages/ripple/src/runtime/internal/client/events.js
```diff
@@ -160,8 +174,6 @@ export function handle_event_propagation(event) {
  * @param {AddEventListenerOptions} [options]
  */
 function create_event(event_name, dom, handler, options = {}) {
-  var block = active_block;
```
leonidaz, line 163: this was not used by the function

trueadm (approved):

.editorconfig
JakeIsMeh, line 1: Should this be a seperate PR?

leonidaz: Yeah but at this point it’s fine

Issue #64, opened 2025-09-04
Add examples


yangshun:
I'm thinking of contributing some examples as a way to get more familiar with the syntax.

Vue and Svelte showcase examples / playground in different ways:
 
- Vue: Separate [examples page](https://vuejs.org/examples/) and [standalone playground page](https://play.vuejs.org/)
- Svelte: Integrated [playground page that allows for switching between examples](https://svelte.dev/playground/)

Points for discussion:

1. Which approach do you wanna go with? Or something entirely different?
2. If going with Svelte approach, should it be added to the existing playground page? If going with the Vue approach, should we wait until the website is up first?

kravallee:
I kinda like both, but the Vue approach feels a bit cleaner/less cluttered to me. The Svelte one is nice for quick switching though.

hatemhosny:
I think a drop down list in the same page can be a simpler option.
This can also be combined with another drop down list for [selecting Ripple version](https://github.com/trueadm/ripple/issues/9#issuecomment-3255863645).

On the other hand, we can add links below examples in the docs to open the code samples in playground.
See: [`getPlaygroundUrl`](https://livecodes.io/docs/sdk/js-ts#getplaygroundurl) method in LiveCodes SDK.
Also these links can be automatically generated from markdown codeblocks. See https://livecodes.io/docs/markdown-to-livecodes

PR #284, opened 2025-09-22
integrate playground in docs website

hatemhosny:
This PR integrates the playground in the docs website:

- Maintains layout (header, footer and website styles)
- Automatically generates share URL
- Uses the same URL (`/playground`). So Previously shared links should keep working.

image

Questions/suggestions: (related: https://github.com/trueadm/ripple/discussions/141)

- Should we add a "share URL" button or just keep auto-generating URLs?
- Should we allow users to change Ripple version? ([we can do that in runtime](https://livecodes.io/?x=id/jn4dg38jmhm))
- The playground can load examples. How do you suggest implementing this? (e.g. drop-down list or side bar) (related #64)
- We can embed small playgrounds inside doc pages for quick demos. 

image

- We can place links to playground under code snippets

image

- LiveCodes can also run React, Svelte and Solid. Do you think we should put comparable code snippets for each in [Comparison to other frameworks](https://www.ripplejs.com/docs/comparison) page? [example](https://racing-bars.hatemhosny.dev/playground/)
- The links on the homepage to [docs](https://github.com/trueadm/ripple/blob/42e8cd249038bee3c85a81f6a1911fc9305bb07b/website/public/index.html#L917) and [playground](https://github.com/trueadm/ripple/blob/42e8cd249038bee3c85a81f6a1911fc9305bb07b/website/public/index.html#L922) currently open in new windows. The is not the expected behaviour for opening links in the same website. Shall I remove ` target="_blank"`?


website/playground.md
```diff
@@ -0,0 +1,18 @@
...
+
+# Playground
+
+<ClientOnly>
+	<div style="min-height: calc(100vh - 200px); margin-bottom: -90px">
+		<Suspense>
+				<Playground :is-main-playground="true" />
+		</Suspense>
+	</div>
+</ClientOnly>
```
leonidaz: We'd want the playground to take up the whole width of the page and we don't need any extra heading or vertical space at the top, it should be flush to the top navigation bar so switching from other site pages, only what's below the top menu changes.

And so, it doesn't need to say Playground as it's pretty obvious and the top right menu with Playground gets highlighted, perhaps we can it more visible but that's for another day.

leonidaz:
First, thank you for keeping on top of this!

> Should we add a "share URL" button or just keep auto-generating URLs?

Yes, but only limited a copy link with no social media

> Should we allow users to change Ripple version? ([we can do that in runtime](https://livecodes.io/?x=id/jn4dg38jmhm))

Yes, I think you showed a dropdown where people can choose, will it be remembered with a share link?

The playground can load examples. How do you suggest implementing this? (e.g. drop-down list or side bar) (related https://github.com/trueadm/ripple/issues/64)

> I think for every code example, we can have two tabs: defaults to `Code` and switching to `Playground` loads the playground, we may limit options here.

> We can embed small playgrounds inside doc pages for quick demos.

as per above.

> LiveCodes can also run React, Svelte and Solid. Do you think we should put comparable code snippets for each in [Comparison to other frameworks](https://www.ripplejs.com/docs/comparison) page? [example](https://racing-bars.hatemhosny.dev/playground/)

not at this time, it's too early for thi. 

> The links on the homepage to [docs](https://github.com/trueadm/ripple/blob/42e8cd249038bee3c85a81f6a1911fc9305bb07b/website/public/index.html#L917) and [playground](https://github.com/trueadm/ripple/blob/42e8cd249038bee3c85a81f6a1911fc9305bb07b/website/public/index.html#L922) currently open in new windows. The is not the expected behaviour for opening links in the same website. Shall I remove  target="_blank"?

Yeah, let's do this

leonidaz (changes requested):

hatemhosny:
Thank you, @leonidaz 

- Playground layout modified as suggested
- Homepage docs and playground links open in the same window
- Removed the old playground HTML page
- Code snippets can now be placed in (Code / Playground) tabs - please see Introduction and Template Syntax pages

To use tabs, the only required change is to place the code block inside a `<Code> ... </Code>` component ([example](https://github.com/hatemhosny/ripple/blob/0f7a7d19bab5ee2d68a11ac661dd7364c9eb2ca5/website/docs/introduction.md?plain=1#L11-L38)). No need to import it. It is auto-registered.

For the playground to know which component to mount as the root, the component needs to be exported as the default.
I'm currently [modifying the code](https://github.com/hatemhosny/ripple/blob/0f7a7d19bab5ee2d68a11ac661dd7364c9eb2ca5/website/.vitepress/components/Code.vue#L29-L40) on the fly to ensure that. Or shall I modify the original code snippets?

If you agree on this direction, I'll start adding the UI elements (share button and drop down version selector).

leonidaz:
the embedded playgrounds with tabs is really cool!

if we can make the playground page go edge to edge.

it's placed in this div.container and on this page it needs to have it's max-width removed and set at 100% and padding removed.

.VPHome .container:has(.main-playground)

and there is top padding that's keeping the playground lower than the top header / menu bar, it should be flash with it directly touching it.

here's what i mean:

image

there is also the document site itself, the content area is pretty narrow by padding / marging and the sidebars take way too much space on desktop.  If we decrease those, we can make a lot more room for the embedded playgrounds.  this is maybe for @JakeIsMeh too handle, unless @hatemhosny you could just do it.  we just got to make sure that lower resolutions are ok.

hatemhosny:
Ok, I'll make it wider.

> and there is top padding that's keeping the playground lower than the top header / menu bar, it should be flash with it directly touching it.

Regarding the space at the top, I was planning to put the share button and version selector above the playground. Otherwise where do you suggest to add these?

> we just got to make sure that lower resolutions are ok.

On mobile devices, the playground can turn to vertical layout 
Screenshot_20250923-051447_Chrome

hatemhosny:
Layout edited to give maximum space for the playground.

Please advise where you want to place the UI elements (share button and version selector)

RATIU5:
Quick question, it appears you are using Starlight for your documentation, but the current documentation pages that are live are using VitePress, which direction are we heading with the docs?

hatemhosny:
> Quick question, it appears you are using Starlight for your documentation, but the current documentation pages that are live are using VitePress, which direction are we heading with the docs?

The docs website uses Vitepress. It was added by this PR: #174

leonidaz:
could we make the playground match the dark / light theme settings?  So, it's not always dark?

hatemhosny:
Added some changes:

- The playground now fills the available space in the page with no scrollbar.
- Added "Copy URL" button above the playground.
- Dark/light mode. This also changes the styles in the result page (dark/light background).

I have been looking at Svelte and Vue playgrounds. These are some suggestions for additional features that they provide:

- Version selector: 
  - should this be a dropdown list? how many versions should we include? (Vue shows last 30)
  - or should it be a text input like Svelte?
- Svelte playground allows users to enable TailwindCSS support and vim mode. These are supported by LiveCodes. Should I do this?

leonidaz:
nice!

> * Version selector:
>   
>   * should this be a dropdown list? how many versions should we include? (Vue shows last 30)

drop down with 30 is good, I'm assuming it's a part of url, so those who need something else can change the url?

>   * or should it be a text input like Svelte?

I think dropdown is a better experience

> * Svelte playground allows users to enable TailwindCSS support and vim mode. These are supported by LiveCodes. Should I do this?

yeah, that'd be nice. I assuming it would be some sort of button for this.

hatemhosny:
Updates:
- Support for Tailwind CSS and Vim mode
- Version selector
- Playgrounds for code blocks in docs can have custom height

image

image



Notes:
- All playgrounds always default to using latest version and with Tailwind disabled.
- Shared URLs preserve version and Tailwind state
- Vim mode config is saved to localStorage. When enabled, any other playground (even inside docs) will have it enabled. Shared URLs, however, do not persist this state (because it is a user preference)
- The version selector only changes the compiler version. The formatter always uses the latest version.

If you agree on that so far, I will go through the rest of code snippets in docs and enable playgrounds for them.

Thank you.

leonidaz:
this is awesome!  Yes, please go for the rest of the snippets. Hopefully they work, you might need to modify them a bit.

my only feedback is if we could replace the the example on the playground page with the full counter example that ripple typically provides and specify some tailwind classes so that when switching to tailwind it looks different.

hatemhosny:
Thank you @leonidaz 
 
> my only feedback is if we could replace the the example on the playground page with the full counter example that ripple typically provides and specify some tailwind classes so that when switching to tailwind it looks different.

Do you mean something [like this](https://ripple-esfysn76a-trueadms-projects.vercel.app/playground#config=code%2FN4IgLglmA2CmIC4QFUB2kawCYAIAKATgPYBWsAxmCADQhawDO5BEADpEaoiDSABawAhlm4AeALawwgnOT6CCDKQF4AOigAqAMQC0ADnU4A9AD5VqCVJmpBktSABuEWAHdWRAmEPlOYWOnsXCCwwPmV6J3JYHSCQvmocCFQoCEFoHSY02GUARgA6AAZDU14%2BMHFoAEEwMEVuaEFUAHN7f28GhgZ7dV5pJoZEAG0AXVpBSggHWABRLCgPbiYWdl5xBQBrAFdWRFAG5s3BJvgkMoreH3R-KiQQAF9aBjAATzhdkH2mw%2BPuck6L3zXbilYL0VAAYUB6G4ACMiFhnjhgDgAGa%2BHQo2wQaDPBA4HSCVisOAZZ5PWDiBIAIWgSXWAFlxgBlMl%2BcRaXwJADkTNgTSIsBwyAAklyEgAlIhwsBEBIAeQAHs9jqgEsgYZt0JsEuDGtICLBoNBuXLWP4cEzGgwxTguQAJQ1TSDkGQAOVgm1gNoYVoysBYKIA3DgcHd7o9mGwbntGl8jicQMtifBaJc-NCkBBxO5PEicLVxutQ6jiOJbUm4FzA%2BZzLAFTmwDh6JjNtBGz5s5xrjhIZq-AQABQASiR5hDcHbRD7OGU%2BYIhYHBSH1dQ46kTanMLgM7nC%2BHM5MOAAAj5pwAqHAAJmXNdXOFEcwcsg6XXUacESX9OFYOgAbDhBgAMgAfQ1GpOGGBApxgT8AJAsCZVQSDoNpVBogYdZnh0ABWAoihAMw7xDUQ%2BEvZ9BE6ew-AVMAdEvBVoFRdE4WgLB1BMYAuV7K4CC5O5RCMUjCJDYjWHIyj1HEHQAGZ2OAAADbiwDxAASYATyndA7nk-ijFYYSRNEfSFIAEU3OBVPUrBzNgbTdP028RPvBDOBwThwVpch1mUYB92UQ8NL7HQdDuDiuR0PiBJc1ADOI6K3IhTzvN8kd-OPU90AAaky0LOMyyKjGi2L70fYq4s2cDV3cpKfL8gKMsbWcClyrlxUYKQ%2BPvQqKsQ4qBNKscuoG1BBtEJ5XlgAy8nfT8CFHIj8zrWi0ggJpUDxKIeJXJy0XQDEsRxPF1EqFg0nUBIfVQBg-QDbbQ0cnB4uAQaQ122iGAgAAvWA8RyCk7pDVhhDmZo8V-VgFRwHJLwhgGcDWAgmiSMHYZe2RNkUDw8XcJJ%2BzhuECHoAg8VQLs4be-bxGxXEcHUXsWC-d0XHO%2BHOCIBggaiO67ge6KED4Igpjm56FphQsmmITUsB0HxoCxnAAGJYAKZXlbh2X5YVy9te50ajHGuBCJ5rgHhAA3GAEKQBgQEYI2WMBrdtkA-gYQh2puHhaFYYgok6DxHdGZ3NieIhxF5Gokn6d4KxOUAhY%2BzhuAKPJLzyAB2X97lNrMG2t4BTZeM084LxgHfeT5vgTQvGEjFZUyhD3w0cf0E64JAABZM9oGUiGgPOQH8QQt2wbg0mgXhxkgKZfk4Bhe5TM3pDAYPfjlpQRDuO4gA)?

My concern is that adding Tailwind classes without having effect might be confusing. The users will start modifying them and then they think it is not working (specially that the switch for enabling Tailwind is hidden inside the menu). What do you think?
Enabling Tailwind adds the CSS resets, so the users will know that it is working without having to add classes.

I suggest leaving the users to add Tailwind classes after enabling it. This is also what Svelte playground does.

hatemhosny:
Updates:

- The playground adds the compiler version to share URLs as the query param `?v=0.2.78`. It can be overriden by `?v=latest`
- Updated the default demo in playground. @leonidaz please check and share any code changes you want there.
- Playgrounds are added to all (runnable) code snippets in docs. I have made some minimal changes in code just enough for it to run or show output. @leonidaz , @JakeIsMeh  would you please review? You can find most docs updates in [this commit](https://github.com/trueadm/ripple/pull/284/commits/5680a737a2e3f714498333e0bf623f2b8320a9a0).

While adding the playgrounds in docs I noticed a few things:
- I was not able to get the [untrack](https://www.ripplejs.com/docs/guide/reactivity#untracking-reactivity) demo to run properly (even after updating to new syntax - clicking the button throws an error)
- The demo of [`createRefKey`](https://www.ripplejs.com/docs/guide/dom-refs#createrefkey) is not reactive
- I found TODO comments [here](https://www.ripplejs.com/docs/guide/syntax#concept-templates-as-lexical-scopes) and [here](https://www.ripplejs.com/docs/guide/syntax#raw-html)

Suggestion:
LiveCodes has an [AI code assistant](https://livecodes.io/docs/features/ai) (powered by [WindSurf](https://windsurf.com/) - formerly Codeium). It is free and does not require any accounts or tokens. I have it enabled as a VS Code extension and it provides quite useful code completions for ripple code. It also works very well with Tailwind styling.
Try it here: https://ripple.livecodes.pages.dev/?template=ripple&enableAI=true
Shall I add that as an option in the playground (like Tailwind and Vim mode)?
image

JakeIsMeh:
Only skimmed through since I'm out rn; Not all code snippets need to have playgrounds — like the ones related to syntax. Good candidates to keep are those that demonstrate reactivity, so that the user can try it out themselves.

hatemhosny:
Updates:

- Added AI assistant to settings menu
- Adjusted playground styles to brand colors
- Removed playground from non-interactive code snippets (mainly in "Syntax" and "Components" pages)
- Some other clean-ups

leonidaz:
> * I was not able to get the [untrack](https://www.ripplejs.com/docs/guide/reactivity#untracking-reactivity) demo to run properly (even after updating to new syntax - clicking the button throws an error)

this is using the old syntax:

can be replaced with:

```jsx
import { untrack } from 'ripple';

export component App(
  let count = track(0);
  let double = track(() => @count * 2);
  let quadruple = track(() => @double * 2);

  effect(() => {
    // This effect will never fire again, as we've untracked the only dependency it has
    console.log(untrack(() => @quadruple));
  })
}
```

> * The demo of [`createRefKey`](https://www.ripplejs.com/docs/guide/dom-refs#createrefkey) is not reactive

It's a bug with composite components, I reported it. #337 

> * I found TODO comments [here](https://www.ripplejs.com/docs/guide/syntax#concept-templates-as-lexical-scopes) and [here](https://www.ripplejs.com/docs/guide/syntax#raw-html)

I don't think we need the todos





> * Added AI assistant to settings menu

This is very cool! good call!

> * Adjusted playground styles to brand colors

👍 

> * Removed playground from non-interactive code snippets (mainly in "Syntax" and "Components" pages)

👍 

> * Some other clean-ups

👍

hatemhosny:
Thank you @leonidaz 

- updated `untrack` code snippet. It works well now. Thank you.
- removed TODOs.
- The `effects` section was duplicated. Removed the second one.

hatemhosny:
Update:
- Added the ability to change playground font size

Suggestion:
Svelte playground provides a list of examples in a dropdown list. Selecting an example loads it in the playground.
I think this provides an interactive way of learning for those who learn by doing rather than by reading.
Do you think we should do something similar?

hatemhosny:
> Suggestion: Svelte playground provides a list of examples in a dropdown list. Selecting an example loads it in the playground. I think this provides an interactive way of learning for those who learn by doing rather than by reading. Do you think we should do something similar?

I have added a demo for examples menu.
If you agree on this approach, I can continue adding examples for the rest of the features.

image

leonidaz:
> I have added a demo for examples menu. If you agree on this approach, I can continue adding examples for the rest of the features.

I think it’s a great idea, let’s to do it. 

Let me know when you wanna merge this, is other changes can be done on a different PR

hatemhosny:
> I think it’s a great idea, let’s to do it.
> 
> Let me know when you wanna merge this, is other changes can be done on a different PR

Thank you, @leonidaz 

I'll add some more examples, so that when it is published it covers the important features (e.g. reactivity).
I'll let you know when I'm ready.

The examples are placed in [`examples.ts` file](https://github.com/hatemhosny/ripple/blob/integrate-playground/website/docs/examples.ts) in `website/docs`. This can be edited later in other PRs as you say.

hatemhosny:
I have added examples that cover most of the snippets in docs.

I think the PR is now ready for review and merge if you think it is ok.

PR #285, opened 2025-09-22
update: favicon on website & docs

RATIU5:
### Description

Adds the new favicon from [the Figma file](https://www.figma.com/design/arPhVlDxkfKVUyTEhJL8Sk/ripple-thoughts) to the website and the docs.

### Reason

Slow progress to making Ripple have a unique brand.

### Notes

Let me know if further modifications are needed.


trueadm:
I was of the understanding that favicons should use `ico` still for security reasons and backwards compat on mobile instead of svg/xml?

RATIU5:
@trueadm I know SVGs are recommended for future-proofing, but I did stumble across [this article](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7) just now and it does recommend keeping an `ico` version for Safari (recently added support for SVGs) and for IE. I'll go ahead and add an `ico` fallback format as well.

trueadm (approved):
Thank you

PR #286, opened 2025-09-22
Adds support for track props split

leonidaz:


leonidaz:
Let’s not merge yet, just thought of something I should try in examples to verify the docs

packages/ripple/src/compiler/phases/3-transform/index.js
```diff
@@ -189,6 +189,8 @@ const visitors = {
     if (!context.state.to_ts && is_ripple_track_call(callee, context)) {
       if (node.arguments.length === 0) {
         node.arguments.push(b.void0);
```
trueadm: Shouldn't you push two void0s here?

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,89 @@ export function derived(fn, block) {
...
+    }
+
+    if (typeof v === 'function') {
+      return derived(v, b);
+    }
+    return tracked(v, b);
+  }
+
+  if (is_tracked || typeof v !== "object" || v === null || is_array(v)) {
+    throw new TypeError('Invalid value: expected a non-tracked object');
```
trueadm, line 303: Side note, we need to extract errors out at prod time, and use error codes instead. Otherwise our bundle size is going to blow up

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,89 @@ export function derived(fn, block) {
...
+  }
+
+  var list;
+
+  if (
+    typeof o !== "object" ||
+    o === null ||
+    is_array(o) ||
```
trueadm: This doesn't seem right?

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,89 @@ export function derived(fn, block) {
...
+      return derived(v, b);
+    }
+    return tracked(v, b);
+  }
+
+  if (is_tracked || typeof v !== "object" || v === null || is_array(v)) {
+    throw new TypeError('Invalid value: expected a non-tracked object');
+  }
+
+  var list;
```
trueadm: why not

`var list = o?.list`

packages/ripple/src/compiler/phases/3-transform/index.js
leonidaz: oh, you wanted it be completely monomorphic... gotcha, yeah, then i can update for sure.

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,89 @@ export function derived(fn, block) {
...
+
+  if (
+    typeof o !== "object" ||
+    o === null ||
+    is_array(o) ||
+    !("split" in o) ||
+    !is_array(list = o.split) ||
+    list.length < 1
+  ) {
+    throw new TypeError(
```
trueadm: I wouldn't bother with this validation

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,89 @@ export function derived(fn, block) {
...
+  var props = Reflect.ownKeys(descriptors);
+  /** @type {Record<string|symbol, Tracked>} */
+  var done = {};
+
+  for (let i = 0, props_i = 0, key, t; i < list.length; i++) {
+    key = list[i];
+
+    if (done[key]) {
+      // in case of duplicate keys just return as requested
+      out[i] = done[key];
```
trueadm: This isn't needed. We can just re-apply duplicates, it's not going to cause issues

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,89 @@ export function derived(fn, block) {
...
+      continue;
+    }
+
+    if (is_tracked_object(v[key])) {
+      t = v[key];
+    } else {
+      t = tracked(undefined, b);
+      delete t.v;
```
trueadm: Why are we deleting a property here? this will cause a deopt on the tracked value

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,89 @@ export function derived(fn, block) {
...
+      define_property(t, 'v', descriptors[key]);
+    }
+
+    out[i] = t;
+    done[key] = t;
+    if ((props_i = props.indexOf(key)) !== -1) {
+      props.splice(props_i, 1);
```
trueadm: This seems unecessary?

packages/ripple/src/runtime/internal/client/runtime.js
leonidaz: yeah, good idea, like autogenerating files with error message codes and messages?

trueadm: Yep, see how React and Svelte do it. I like how Svelte does it though, they go into a markdown file and get built to be imported.

packages/ripple/src/compiler/phases/3-transform/index.js
leonidaz: duh... sorry yeah, it had to be there.

packages/ripple/src/runtime/internal/client/runtime.js
leonidaz: how do you mean?  just trying to eliminate an object that could be an array.

but doesn't matter, i'll remove this validation as per your comment below.  I wasn't sure if we want to do in exchange of increasing size and processing time.

packages/ripple/src/runtime/internal/client/runtime.js
leonidaz: I need to know which props are left for the `rest`. I can just delete the prop on descriptors and then create the array for the `rest`.

packages/ripple/src/runtime/internal/client/runtime.js
leonidaz: overthought it, i was thinking of `configurable` set to false, and i wouldn't be able to overwrite, but that's never going to be the case as we set it ourselves in tracked.

packages/ripple/src/runtime/internal/client/runtime.js
leonidaz: I'm assuming you want to keep this validation vs the other one that I removed.

packages/ripple/src/runtime/internal/client/runtime.js
trueadm: I just feel like the splice here is far too expensive, maybe re-work it? Also do we need to use Reflect.ownKeys?

leonidaz: yeah, i agree , originally also thought that but given that arrays would be small and and this split operation is not going to be used in a loop or something.

but I removed it.  deleting keys is much faster and then getting the array to loop over.

Yea, we need Reflect.ownKeys because they may contain Symbols that we need to keep, like ref.  The only other way is to use two methods which would be slower.

packages/ripple/src/runtime/internal/client/runtime.js
trueadm: We can keep this one :)

packages/ripple/tsconfig.json
```diff
@@ -11,7 +11,7 @@
 		"allowSyntheticDefaultImports": true,
 		"verbatimModuleSyntax": true,
 		"types": ["node"],
-    "jsx": "preserve",
+    	"jsx": "preserve",
```
trueadm: whitespace?

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -278,6 +279,62 @@ export function derived(fn, block) {
...
+    key = list[i];
+
+    if (is_tracked_object(v[key])) {
+      t = v[key];
+    } else {
+      t = define_property(tracked(undefined, b), 'v', descriptors[key]);
+    }
+
+    out[i] = t;
+    delete descriptors[key]
```
trueadm: It would be nice to not `delete` here still. Can we not `null` the field?

packages/ripple/tsconfig.json
leonidaz: ah, it had spaces, ok, will replace by tabs.

packages/ripple/src/runtime/internal/client/runtime.js
leonidaz: ok, pushed the changes. 

l set it to null, as there is no deopt,  tweaked the types and just skipped nulls it in the loop.  The tradeoff is more loop iterations.

PR #287, opened 2025-09-22
Update readme assets

RATIU5:
### Description

Replaces the old logos with updated ones. The PNG images have been scaled and sized to look the best on GitHub. The color for the light-mode logo (#3989EB) was chosen for it's increased contrast versus the dark-mode logo (#3E95FF) to meet WCAG AA compliance. Colors were chosen from the [Ripple design guide on Figma](https://www.figma.com/design/arPhVlDxkfKVUyTEhJL8Sk/ripple-thoughts).


trueadm (approved):
I love it. Let’s get feedback from the team too

leonidaz:
I would make it smaller like about 72% of what's it now. 730px wide max on desktop but not sure if there are options for mobile as if it scales like this it will be too small.

Also, I think we can use the same one for light mode as for the dark.  I like that it's bright.

RATIU5:
@leonidaz  Sorry for the many commits, how does this size look to you now?

leonidaz:
no problem!  the image size is going down but not the width.

it's still 1800  wide, i just downloaded to check, making sure i'm not getting something cached.

but the bright on white is good.

RATIU5:
@leonidaz  Oh, I was under the impression you wanted the image to appear visually smaller, so I would need to increase the width of the blank space so that it would appear smaller. You are saying that the actual image size as a whole should go down? I can do that. Before I do, do you want the original visual logo size in comparison to the rest of the rendered README, or do you want the other size I had originally back (just shrunk to 72%)?

leonidaz:
the original quality was good, yeah I was concerned with the actual visual size, it appeared to be way too big on desktop.

I just tried it the browser and for max-width I set to 72% from 100% from the 1800 wide image.  it seemed to scale it proportionally.  but for mobile it looked too small when scaled down and i'm not sure if there is a separate option for smaller screen resolutions, in which case we can still make less than 1800 px by maybe to 85% or something like that.

RATIU5:
I think I was able to get it working the way you described, let me know if you want more changes.

leonidaz:
ok, I think it's good now, thank you!

leonidaz:
i'm just going to merge this, we can always adjust later.

PR #292, opened 2025-09-23
Split LSP from VSCode Plugin

JakeIsMeh:
todo:
- [ ] remove LS from vsc plugin
- [ ] claim package on npm


leonidaz:
just for clarity, what are we trying to accomplish with this?

kuechlerm:
@leonidaz my guess would be that we could use the LSP in other Editors like Neovim. 
I dont use VS Code, so that would be really helpful.

Yusyuriv:
Hey, sorry for interfering with your PR, but I made a PR that does the same thing: #423. My original intention was to write a native plugin for IntelliJ, but extracting the LSP and making it work in the IDE turned out much simpler, so I did that.

PR #294, opened 2025-09-23
Update `transporting reactivity` section in readme.

sreekar339339:
Tracked values don't require additional level of boxing via arrays or objects anymore like they did with the previous iteration of reactive values..


trueadm:
Thanks

PR #327, opened 2025-09-26
fix: split should return undefined for non-existent props

leonidaz:


packages/ripple/tests/client/basic.test.ripple
```diff
@@ -379,8 +379,7 @@ describe('basic client', () => {
 		expect(container.querySelector('.count').textContent).toBe('0');
 	});
 
-	// TODO
-	it.skip('renders multiple reactive lexical blocks with complexity', () => {
+	it('renders multiple reactive lexical blocks with complexity', () => {
```
leonidaz, line 382: i changed this test to actually test for the split as with just tracked `count` passed in, it would work regardless

trueadm (approved):

PR #338, opened 2025-09-27
feat: TrackedObject and shorthand reactive #{...} object

leonidaz:
- Adds a shallowly reactive TrackedObject including the shorthand `#{...}` notation
- Moves the proxy code from array.js into its own file `proxy.js`
- The proxy in `proxy.js` works with both TrackedArray and TrackedObject


packages/ripple/src/runtime/array.js
```diff
@@ -55,246 +53,15 @@ TrackedArray.fromAsync = async function (arrayLike, mapFn, thisArg) {
   var elements = mapFn
     ? await Array.fromAsync(arrayLike, mapFn, thisArg)
     : await Array.fromAsync(arrayLike);
-  return proxy({ elements, block, from_static: true });
+  return array_proxy({ elements, block, from_static: true });
 };
 
 /**
  * @template T
- * @param {{
```
leonidaz, line 63: most of the array.js got move into proxy.js

the proxy.js is used by both TrackedArray and TrackedObject

trueadm (approved):

PR #344, opened 2025-09-28
Fix spread events

anubra266:


trueadm (approved):
Looks good, I wonder if we need to allow casing in cases for custom events? I guess we can get around to that another time.

PR #346, opened 2025-09-28
Fix some tests failing on Windows because of backslashes in paths

Yusyuriv:
These 3 tests were failing on Windows because they expected the paths to contain forward slashes, but on Windows `getLocalTemplatePath()` returns paths with backwards slashes. I added a function to normalize those paths for tests, now they successfully path both on Linux and Windows.


trueadm (approved):

trueadm:
Thanks!

PR #347, opened 2025-09-28
fix: Fix some tests failing on older versions of Node

Yusyuriv:
The tests were failing because older versions of Node don't have `Array#toSorted` and `Array#toSpliced`, which caused the code to call `context.skip()`, but `context` was `undefined`. I added the `context` argument to those tests, and they now successfully skip when executed on old Node.


trueadm:
Thank you

PR #348, opened 2025-09-28
fix: documentation quickstart

WebEferen:
Fixed documentation quickstart guide to properly follow the create-ripple logic.


JakeIsMeh:
Is there a reason for this change? `npm create <name>` calls `create-<name>` by convention

WebEferen:
@JakeIsMeh good point, it seemed like the `create XX` command had some issues over this weekend and it was wrongly pointing to the `create create-ripple` instead of `ripple`.

For now it works correctly - closing the PR (if the issue comes back, I'd just update it)

PR #355, opened 2025-09-28
improve docs

JakeIsMeh:
- Reworded event prop binding to be more specific
- Documented `<head>`
- Documented unscoped styles (or lack thereof)
- Normalized prose to 80 chars long

### DIFF
--- a/website/docs/guide/events.md
+++ b/website/docs/guide/events.md
@@ -7,7 +7,7 @@ title: Event Binding in Ripple
 Ripple deviates slightly from standard JS in terms of event names. Like most
 frameworks except for React, we rely on the native event system of the browser.
 However, Ripple doesn't use the same all-lowercase props/attributes to bind
-events -- a camelCase convention is used instead, shown below:
+events — a camelCase convention is used instead, shown below:
 
 - `onclick` -> `onClick`
 - `onpointermove` -> `onPointerMove`


website/docs/guide/components.md
```diff
@@ -21,9 +21,9 @@ within `effect()` to ensure they only run when intended.
 
 ## Children
 
-Use `children` prop and then use it in the form of `<children />` for component composition.
-
-When you pass in children to a component, it gets implicitly passed as the `children` prop, in the form of a component.
+Use `children` prop and then use it in the form of `<children />` for component
```
leonidaz: i think this sounds weird with two `use` close by.

website/docs/guide/components.md
```diff
@@ -21,9 +21,9 @@ within `effect()` to ensure they only run when intended.
...
+composition. When you pass in children to a component, it gets implicitly passed
+as the `children` prop, in the form of a component.
```
leonidaz: we should probably mention, doesn't have to be this time, that you can just assign a component to the childrent prop:

`<Child children={MyComponent} />`

website/docs/guide/styling.md
```diff
@@ -45,3 +48,8 @@ let count = track(3);
 <div class={['foo', {bar: @count > 2}, @count > 3 && 'bat']}></div>
 // becomes: class="foo bar"
 ```
+
+## Global Styles
+
+Currently, there is no way to opt out of scoped styles other than placing your
+styles in a style tag within index.html's head, or importing the CSS with Vite.
```
leonidaz: we should just say that the support for `:global` is coming soon, otherwise it sounds too drastic.

leonidaz (changes requested):

website/docs/guide/control-flow.md
```diff
@@ -4,31 +4,11 @@ title: Control flow in Ripple
...
-
-  {str}
-</div>
-```
-
-Note that strings inside the template need to be inside `{"string"}`, you can't do `<div>hello</div>` as Ripple
-has no idea if `hello` is a string or maybe some JavaScript code that needs evaluating, so just ensure you wrap them
-in curly braces. This shouldn't be an issue in the real-world anyway, as you'll likely use an i18n library that means
-using JavaScript expressions regardless.
-
```
leonidaz, line 27: is this explained on some other page?  I think it's useful to have this section, maybe not under Control Flow but maybe but something like General or Templating.

website/docs/guide/events.md
```diff
@@ -4,13 +4,15 @@ title: Event Binding in Ripple
 
 # Event Binding with Props
 
-Like React, Ripple deviates from the standard with event handlers. Here, events
-are camelCased props that start with 'on', unlike standard HTML.
+Ripple deviates slightly from the platform in terms of event binding. Like most
```
leonidaz: this is weird `from the platform`, should be `from the standard` and it's not in terms of event binding but purely based on the event naming with camelCase.

`slightly` also a strange choice, i'd just omit it and say `in terms of event naming...`

website/docs/guide/reactivity.md
```diff
@@ -153,20 +180,29 @@ export component App() {
...
 
 ::: info Note
+<<<<<<< HEAD
 Make sure the resulting `rest`, if it's going to be spread onto a dom element, does not contain `Tracked` values. Otherwise, you'd be spreading not the actual values but the boxed ones, which are objects that will appear as `[object Object]` on the dom element.
+=======
+Make sure the resulting `rest`, if it's going to be spread onto a dom element,
+does not contain `Tracked` values. Otherwise, you'd be spreading not the actual
+values but the boxed ones, which are objects that will appear as
+`[Object object]` on the dom element.
+>>>>>>> 4ae1575 (make raw files for docs 80ch max width)
```
leonidaz: oops

website/docs/guide/styling.md
```diff
@@ -75,3 +78,8 @@ Both examples above will render the same inline styles, however, it's recommende
 ::: info
 When passing an object to the `style` attribute, you can use either camelCase or kebab-case for CSS property names.
 :::
+## Global Styles
+
+Currently, there is no way to opt out of scoped styles (though the team is
+working on it!) other than placing your styles in a style tag within index.html's
+head, or importing the CSS with Vite.
```
leonidaz: still sounds kind of harsh, let's do the following:

```suggestion
At the moment, marking styles as global with `:global` inside components isn’t
supported yet (though the team is working on adding it!). For now, you can place
your styles in a `<style>` tag inside `index.html`’s `<head>`, or import your CSS
with Vite as a workaround.
```

leonidaz (changes requested):

website/docs/guide/control-flow.md
JakeIsMeh: it's explained on the syntax page (docs/guide/syntax)

website/docs/guide/events.md
JakeIsMeh: "from the platform" is referring to "use the platform", a term you might have seen floating around the Svelte sphere of influence, and also the the docs for [SvelteKit](https://svelte.dev/docs/kit/web-standards). The use of "slightly" is to prime readers for the next sentence. (Attempting to prevent React migrators from thinking that Ripple has a full on synthetic event system)

website/docs/guide/events.md
```diff
@@ -4,13 +4,15 @@ title: Event Binding in Ripple
 
 # Event Binding with Props
 
-Like React, Ripple deviates from the standard with event handlers. Here, events
-are camelCased props that start with 'on', unlike standard HTML.
+Ripple deviates slightly from standard JS in terms of event names. Like most
+frameworks except for React, we rely on the native event system of the browser.
+However, Ripple doesn't use the same all-lowercase props/attributes to bind
+events -- a camelCase convention is used instead, shown below:
```
leonidaz: nit to to get rid of the em dash completely and rephrase a bit simpler.

otherwise, need to replace 2 double dashes with one em dash —

```suggestion
Like most frameworks except for React, Ripple relies on the browser’s
native event system. However, Ripple doesn’t use the same all-lowercase
props/attributes to bind events. Instead, it follows a camelCase convention,
as shown below:
```

leonidaz (approved):
just one nit otherwise 👍

PR #359, opened 2025-09-28
docs: document HTML directive

JakeIsMeh:


README.md
```diff
@@ -747,6 +711,24 @@ component Input({ id, value, ...rest }) {
...
+By default, all text nodes in Ripple are escaped to prevent unintended script
+injections. If you'd like to render trusted HTML onto your page, you can use the
+HTML directive to opt-out:
+
+````ripple
+export component RawContent() {
+	let source = ```
+# My Markdown Document
+Hi! I like JS and Ripple.
+```
```
leonidaz: wonder if we should just use a straight html example, and also use `export component App` so someone could copy this and run it.

leonidaz (approved):
👍

PR #362, opened 2025-09-28
feat: Allow using objects for style attribute

### LINKED ISSUE #323: [Feature]: Add support for dynamic inline styles
### Feature Category

Syntax/Language Improvement

### Problem Statement

# Feature Request: Dynamic inline styles

Hi 👋 first of all, thanks for the awesome work on RippleJS!  
While experimenting, I noticed that styling based on conditions is not very ergonomic right now.  

Currently the only way I found is using `track`, for example:

`let color = track(() => @count > 10 ? "red" : "blue")`

This works, but feels a bit verbose when you want to handle multiple style properties or simple conditions.

The problem
In other reactive frameworks (like Svelte or Solid), it’s very natural to bind styles directly, e.g.:

`<div style={{ width: variable1, color: variable2 }}></div>`

This makes dynamic styles really easy to write and also very familiar for people coming from those ecosystems.

### Proposed Solution

It would be great if RippleJS could support something like:

`<div style={{ color: @color, width: @width }}></div>`

or maybe a shorthand syntax like:

`<div style:{@color}></div>`
That way, we could directly bind reactive values to inline styles without always having to create extra track variables.

### Alternatives Considered

_No response_

### Examples & References

_No response_

### Checklist

- [ ] I'm willing to help implement this feature

Yusyuriv:
Closes #323.

## What it is
This PR allows specifying an object as the value of the `style` attribute. For example:
```jsx
let color = track('red');

function toggleColor() {
  @color = @color === 'red' ? 'green' : 'red';
}

// This produces:
// style="color: red; font-weight: bold"
<div style={{
  color: @color,
  fontWeight: 'bold',
}}>
  {`Dynamic styling`}
</div>

<button onClick={toggleColor}>
  {`Toggle color`}
</button>
```

## Testing
I added 4 new tests for the new functionality and then ran `pnpm test`, all tests successfully passed.


packages/ripple/src/runtime/internal/client/render.js
```diff
@@ -85,7 +86,9 @@ export function set_attribute(element, attribute, value) {
 
   if (value == null) {
     element.removeAttribute(attribute);
-  } else if (typeof value !== 'string' && get_setters(element).includes(attribute)) {
+  } else if (attribute === 'style') {
+		element.setAttribute(attribute, normalize_style(value));
```
leonidaz: I think we probably want to `.setProperty` vs doing the whole thing via `.setAttribute`.  And we can do the normalization as part of this function's loops. And in the `if` test if it's a string vs object and use `setAttribute`.

e.g.

```js
function camelToKebab(prop) {
  if (prop.startsWith('--')) {
    return prop
  }
  return prop.replace(/[A-Z]/g, m => "-" + m.toLowerCase());
}

function applyStyles(el, newStyles) {
  var style = el.style;
  var normalized = {};

  for (var [prop, value] of Object.entries(newStyles)) {
    var cssProp = camelToKebab(prop);
    value = String(value);

    if (style.getPropertyValue(cssProp) !== value) {
      style.setProperty(cssProp, value);
    }

    normalized[cssProp] = value;
  }

  for (var i = style.length - 1; i >= 0; i--) {
    const prop = style[i];
    if (!(prop in normalized)) {
      style.removeProperty(prop);
    }
  }
}
```

Yusyuriv: Done!

packages/ripple/src/utils/normalize_css_property_name.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Takes a camelCased string and returns a hyphenated string
+ * @param {string} str
+ * @returns {string}
+ * @example
+ * normalize_css_property_name('backgroundColor') // 'background-color'
+ */
+export function normalize_css_property_name(str) {
```
trueadm, line 11: I wonder if we should cache them as we go to avoid doing this expensive operation each time?

leonidaz: You mean a map of camel to kebab lookups? Maybe even have a cache upfront on all known css props? I was also thinking of requiring kebab case to be provided in the first place.

Yusyuriv: Done!

Yusyuriv: If you mean requiring the use of kebab case in the `style` attribute object, in my opinion that's rather inconvenient because it'll force developers to wrap a lot of property keys into a string, which gets pretty annoying:
```jsx
<div style={{
  backgroundColor: @bg,
  borderRadius: @radius,
  fontWeight: @fontWeight,
}} />

// vs
<div style={{
  'background-color': @bg,
  'border-radius': @radius,
  'font-weight': @fontWeight,
}} />
```

packages/ripple/src/utils/normalize_css_property_name.js
```diff
@@ -0,0 +1,20 @@
...
+	if (str.startsWith('--')) return str;
+	if (normalized_properties_cache.has(str)) {
```
leonidaz, line 13: we can just use .get and check for undefined since the map values cannot be undefined

packages/ripple/src/utils/normalize_css_property_name.js
leonidaz: Yeah, it’s a bit more inconvenient but at least for me a lot more readable and no lookups / normalizations. With the cache, it should be at least faster. But, yeah, I don’t think people will like it.

packages/ripple/src/utils/normalize_css_property_name.js
Yusyuriv: I thought about that, but then we'd have to call .set every time, even if the value didn't change:
```js
export function normalize_css_property_name(str) {
	if (str.startsWith('--')) return str;
	
	const normalized_result =
		normalized_properties_cache.get(str)
		?? str.replace(/[A-Z]/g, m => '-' + m.toLowerCase());
	normalized_properties_cache.set(str, normalized_result);

	return normalized_result;
}
```
[I ran this](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBdGxqUC02TzB6VjZDakwtVVFhSmIiLCJjb2RlIjoiLyoqIEB0eXBlIHtNYXA8c3RyaW5nLCBzdHJpbmc-fSAqL1xuY29uc3Qgbm9ybWFsaXplZF9wcm9wZXJ0aWVzX2NhY2hlID0gbmV3IE1hcCgpO1xuXG4vKipcbiAqIFRha2VzIGEgY2FtZWxDYXNlZCBzdHJpbmcgYW5kIHJldHVybnMgYSBoeXBoZW5hdGVkIHN0cmluZ1xuICogQHBhcmFtIHtzdHJpbmd9IHN0clxuICogQHJldHVybnMge3N0cmluZ31cbiAqIEBleGFtcGxlXG4gKiBub3JtYWxpemVfY3NzX3Byb3BlcnR5X25hbWUoJ2JhY2tncm91bmRDb2xvcicpIC8vICdiYWNrZ3JvdW5kLWNvbG9yJ1xuICovXG5mdW5jdGlvbiBub3JtYWxpemVfY3NzX3Byb3BlcnR5X25hbWUoc3RyKSB7XG5cdGlmIChzdHIuc3RhcnRzV2l0aCgnLS0nKSkgcmV0dXJuIHN0cjtcblxuXHRjb25zdCBub3JtYWxpemVkX3Jlc3VsdCA9XG5cdFx0bm9ybWFsaXplZF9wcm9wZXJ0aWVzX2NhY2hlLmdldChzdHIpXG5cdFx0Pz8gc3RyLnJlcGxhY2UoL1tBLVpdL2csIG0gPT4gJy0nICsgbS50b0xvd2VyQ2FzZSgpKTtcblx0bm9ybWFsaXplZF9wcm9wZXJ0aWVzX2NhY2hlLnNldChzdHIsIG5vcm1hbGl6ZWRfcmVzdWx0KTtcblxuXHRyZXR1cm4gbm9ybWFsaXplZF9yZXN1bHQ7XG59XG5cbkRBVEEubWFwKG5vcm1hbGl6ZV9jc3NfcHJvcGVydHlfbmFtZSk7IiwibmFtZSI6IlJlcGVhdCBzZXRzIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6Ikd2NlhodFV6cmc1UXpvbTI0Z25yNyIsImNvZGUiOiIvKiogQHR5cGUge01hcDxzdHJpbmcsIHN0cmluZz59ICovXG5jb25zdCBub3JtYWxpemVkX3Byb3BlcnRpZXNfY2FjaGUgPSBuZXcgTWFwKCk7XG5cbi8qKlxuICogVGFrZXMgYSBjYW1lbENhc2VkIHN0cmluZyBhbmQgcmV0dXJucyBhIGh5cGhlbmF0ZWQgc3RyaW5nXG4gKiBAcGFyYW0ge3N0cmluZ30gc3RyXG4gKiBAcmV0dXJucyB7c3RyaW5nfVxuICogQGV4YW1wbGVcbiAqIG5vcm1hbGl6ZV9jc3NfcHJvcGVydHlfbmFtZSgnYmFja2dyb3VuZENvbG9yJykgLy8gJ2JhY2tncm91bmQtY29sb3InXG4gKi9cbmZ1bmN0aW9uIG5vcm1hbGl6ZV9jc3NfcHJvcGVydHlfbmFtZShzdHIpIHtcblx0aWYgKHN0ci5zdGFydHNXaXRoKCctLScpKSByZXR1cm4gc3RyO1xuXHRpZiAobm9ybWFsaXplZF9wcm9wZXJ0aWVzX2NhY2hlLmhhcyhzdHIpKSB7XG5cdFx0cmV0dXJuIC8qKiBAdHlwZSB7c3RyaW5nfSAqLyAobm9ybWFsaXplZF9wcm9wZXJ0aWVzX2NhY2hlLmdldChzdHIpKTtcblx0fVxuXHRjb25zdCBub3JtYWxpemVkX3Jlc3VsdCA9IHN0ci5yZXBsYWNlKC9bQS1aXS9nLCBtID0-ICctJyArIG0udG9Mb3dlckNhc2UoKSk7XG5cdG5vcm1hbGl6ZWRfcHJvcGVydGllc19jYWNoZS5zZXQoc3RyLCBub3JtYWxpemVkX3Jlc3VsdCk7XG5cblx0cmV0dXJuIG5vcm1hbGl6ZWRfcmVzdWx0O1xufVxuXG5EQVRBLm1hcChub3JtYWxpemVfY3NzX3Byb3BlcnR5X25hbWUpOyIsIm5hbWUiOiJoYXMvZ2V0IiwiZGVwZW5kZW5jaWVzIjpbXX1dLCJjb25maWciOnsibmFtZSI6IkJhc2ljIGV4YW1wbGUiLCJwYXJhbGxlbCI6dHJ1ZSwiZ2xvYmFsVGVzdENvbmZpZyI6eyJkZXBlbmRlbmNpZXMiOltdfSwiZGF0YUNvZGUiOiJjb25zdCBrZXlzID0gW1wiYWNjZW50Q29sb3JcIixcblwiYWxpZ25Db250ZW50XCIsXG5cImFsaWduSXRlbXNcIixcblwiYWxpZ25TZWxmXCIsXG5cImFsaWdubWVudEJhc2VsaW5lXCIsXG5cImFsbFwiLFxuXCJhbmltYXRpb25cIixcblwiYW5pbWF0aW9uQ29tcG9zaXRpb25cIixcblwiYW5pbWF0aW9uRGVsYXlcIixcblwiYW5pbWF0aW9uRGlyZWN0aW9uXCIsXG5cImFuaW1hdGlvbkR1cmF0aW9uXCIsXG5cImFuaW1hdGlvbkZpbGxNb2RlXCIsXG5cImFuaW1hdGlvbkl0ZXJhdGlvbkNvdW50XCIsXG5cImFuaW1hdGlvbk5hbWVcIixcblwiYW5pbWF0aW9uUGxheVN0YXRlXCIsXG5cImFuaW1hdGlvblRpbWluZ0Z1bmN0aW9uXCIsXG5cImFwcGVhcmFuY2VcIixcblwiYXNwZWN0UmF0aW9cIixcblwiYmFja2Ryb3BGaWx0ZXJcIixcblwiYmFja2ZhY2VWaXNpYmlsaXR5XCIsXG5cImJhY2tncm91bmRcIixcblwiYmFja2dyb3VuZEF0dGFjaG1lbnRcIixcblwiYmFja2dyb3VuZEJsZW5kTW9kZVwiLFxuXCJiYWNrZ3JvdW5kQ2xpcFwiLFxuXCJiYWNrZ3JvdW5kQ29sb3JcIixcblwiYmFja2dyb3VuZEltYWdlXCIsXG5cImJhY2tncm91bmRPcmlnaW5cIixcblwiYmFja2dyb3VuZFBvc2l0aW9uXCIsXG5cImJhY2tncm91bmRQb3NpdGlvblhcIixcblwiYmFja2dyb3VuZFBvc2l0aW9uWVwiLFxuXCJiYWNrZ3JvdW5kUmVwZWF0XCIsXG5cImJhY2tncm91bmRTaXplXCIsXG5cImJhc2VsaW5lU2hpZnRcIixcblwiYmFzZWxpbmVTb3VyY2VcIixcblwiYmxvY2tTaXplXCIsXG5cImJvcmRlclwiLFxuXCJib3JkZXJCbG9ja1wiLFxuXCJib3JkZXJCbG9ja0NvbG9yXCIsXG5cImJvcmRlckJsb2NrRW5kXCIsXG5cImJvcmRlckJsb2NrRW5kQ29sb3JcIixcblwiYm9yZGVyQmxvY2tFbmRTdHlsZVwiLFxuXCJib3JkZXJCbG9ja0VuZFdpZHRoXCIsXG5cImJvcmRlckJsb2NrU3RhcnRcIixcblwiYm9yZGVyQmxvY2tTdGFydENvbG9yXCIsXG5cImJvcmRlckJsb2NrU3RhcnRTdHlsZVwiLFxuXCJib3JkZXJCbG9ja1N0YXJ0V2lkdGhcIixcblwiYm9yZGVyQmxvY2tTdHlsZVwiLFxuXCJib3JkZXJCbG9ja1dpZHRoXCIsXG5cImJvcmRlckJvdHRvbVwiLFxuXCJib3JkZXJCb3R0b21Db2xvclwiLFxuXCJib3JkZXJCb3R0b21MZWZ0UmFkaXVzXCIsXG5cImJvcmRlckJvdHRvbVJpZ2h0UmFkaXVzXCIsXG5cImJvcmRlckJvdHRvbVN0eWxlXCIsXG5cImJvcmRlckJvdHRvbVdpZHRoXCIsXG5cImJvcmRlckNvbGxhcHNlXCIsXG5cImJvcmRlckNvbG9yXCIsXG5cImJvcmRlckVuZEVuZFJhZGl1c1wiLFxuXCJib3JkZXJFbmRTdGFydFJhZGl1c1wiLFxuXCJib3JkZXJJbWFnZVwiLFxuXCJib3JkZXJJbWFnZU91dHNldFwiLFxuXCJib3JkZXJJbWFnZVJlcGVhdFwiLFxuXCJib3JkZXJJbWFnZVNsaWNlXCIsXG5cImJvcmRlckltYWdlU291cmNlXCIsXG5cImJvcmRlckltYWdlV2lkdGhcIixcblwiYm9yZGVySW5saW5lXCIsXG5cImJvcmRlcklubGluZUNvbG9yXCIsXG5cImJvcmRlcklubGluZUVuZFwiLFxuXCJib3JkZXJJbmxpbmVFbmRDb2xvclwiLFxuXCJib3JkZXJJbmxpbmVFbmRTdHlsZVwiLFxuXCJib3JkZXJJbmxpbmVFbmRXaWR0aFwiLFxuXCJib3JkZXJJbmxpbmVTdGFydFwiLFxuXCJib3JkZXJJbmxpbmVTdGFydENvbG9yXCIsXG5cImJvcmRlcklubGluZVN0YXJ0U3R5bGVcIixcblwiYm9yZGVySW5saW5lU3RhcnRXaWR0aFwiLFxuXCJib3JkZXJJbmxpbmVTdHlsZVwiLFxuXCJib3JkZXJJbmxpbmVXaWR0aFwiLFxuXCJib3JkZXJMZWZ0XCIsXG5cImJvcmRlckxlZnRDb2xvclwiLFxuXCJib3JkZXJMZWZ0U3R5bGVcIixcblwiYm9yZGVyTGVmdFdpZHRoXCIsXG5cImJvcmRlclJhZGl1c1wiLFxuXCJib3JkZXJSaWdodFwiLFxuXCJib3JkZXJSaWdodENvbG9yXCIsXG5cImJvcmRlclJpZ2h0U3R5bGVcIixcblwiYm9yZGVyUmlnaHRXaWR0aFwiLFxuXCJib3JkZXJTcGFjaW5nXCIsXG5cImJvcmRlclN0YXJ0RW5kUmFkaXVzXCIsXG5cImJvcmRlclN0YXJ0U3RhcnRSYWRpdXNcIixcblwiYm9yZGVyU3R5bGVcIixcblwiYm9yZGVyVG9wXCIsXG5cImJvcmRlclRvcENvbG9yXCIsXG5cImJvcmRlclRvcExlZnRSYWRpdXNcIixcblwiYm9yZGVyVG9wUmlnaHRSYWRpdXNcIixcblwiYm9yZGVyVG9wU3R5bGVcIixcblwiYm9yZGVyVG9wV2lkdGhcIixcblwiYm9yZGVyV2lkdGhcIixcblwiYm90dG9tXCIsXG5cImJveERlY29yYXRpb25CcmVha1wiLFxuXCJib3hTaGFkb3dcIixcblwiYm94U2l6aW5nXCIsXG5cImJyZWFrQWZ0ZXJcIixcblwiYnJlYWtCZWZvcmVcIixcblwiYnJlYWtJbnNpZGVcIixcblwiY2FwdGlvblNpZGVcIixcblwiY2FyZXRDb2xvclwiLFxuXCJjbGVhclwiLFxuXCJjbGlwXCIsXG5cImNsaXBQYXRoXCIsXG5cImNsaXBSdWxlXCIsXG5cImNvbG9yXCIsXG5cImNvbG9ySW50ZXJwb2xhdGlvblwiLFxuXCJjb2xvckludGVycG9sYXRpb25GaWx0ZXJzXCIsXG5cImNvbG9yU2NoZW1lXCIsXG5cImNvbHVtbkNvdW50XCIsXG5cImNvbHVtbkZpbGxcIixcblwiY29sdW1uR2FwXCIsXG5cImNvbHVtblJ1bGVcIixcblwiY29sdW1uUnVsZUNvbG9yXCIsXG5cImNvbHVtblJ1bGVTdHlsZVwiLFxuXCJjb2x1bW5SdWxlV2lkdGhcIixcblwiY29sdW1uU3BhblwiLFxuXCJjb2x1bW5XaWR0aFwiLFxuXCJjb2x1bW5zXCIsXG5cImNvbnRhaW5cIixcblwiY29udGFpbkludHJpbnNpY0Jsb2NrU2l6ZVwiLFxuXCJjb250YWluSW50cmluc2ljSGVpZ2h0XCIsXG5cImNvbnRhaW5JbnRyaW5zaWNJbmxpbmVTaXplXCIsXG5cImNvbnRhaW5JbnRyaW5zaWNTaXplXCIsXG5cImNvbnRhaW5JbnRyaW5zaWNXaWR0aFwiLFxuXCJjb250YWluZXJcIixcblwiY29udGFpbmVyTmFtZVwiLFxuXCJjb250YWluZXJUeXBlXCIsXG5cImNvbnRlbnRcIixcblwiY29udGVudFZpc2liaWxpdHlcIixcblwiY291bnRlckluY3JlbWVudFwiLFxuXCJjb3VudGVyUmVzZXRcIixcblwiY291bnRlclNldFwiLFxuXCJjc3NGbG9hdFwiLFxuXCJjc3NUZXh0XCIsXG5cImN1cnNvclwiLFxuXCJjeFwiLFxuXCJjeVwiLFxuXCJkXCIsXG5cImRpcmVjdGlvblwiLFxuXCJkaXNwbGF5XCIsXG5cImRvbWluYW50QmFzZWxpbmVcIixcblwiZW1wdHlDZWxsc1wiLFxuXCJmaWxsXCIsXG5cImZpbGxPcGFjaXR5XCIsXG5cImZpbGxSdWxlXCIsXG5cImZpbHRlclwiLFxuXCJmbGV4XCIsXG5cImZsZXhCYXNpc1wiLFxuXCJmbGV4RGlyZWN0aW9uXCIsXG5cImZsZXhGbG93XCIsXG5cImZsZXhHcm93XCIsXG5cImZsZXhTaHJpbmtcIixcblwiZmxleFdyYXBcIixcblwiZmxvYXRcIixcblwiZmxvb2RDb2xvclwiLFxuXCJmbG9vZE9wYWNpdHlcIixcblwiZm9udFwiLFxuXCJmb250RmFtaWx5XCIsXG5cImZvbnRGZWF0dXJlU2V0dGluZ3NcIixcblwiZm9udEtlcm5pbmdcIixcblwiZm9udE9wdGljYWxTaXppbmdcIixcblwiZm9udFBhbGV0dGVcIixcblwiZm9udFNpemVcIixcblwiZm9udFNpemVBZGp1c3RcIixcblwiZm9udFN0cmV0Y2hcIixcblwiZm9udFN0eWxlXCIsXG5cImZvbnRTeW50aGVzaXNcIixcblwiZm9udFN5bnRoZXNpc1NtYWxsQ2Fwc1wiLFxuXCJmb250U3ludGhlc2lzU3R5bGVcIixcblwiZm9udFN5bnRoZXNpc1dlaWdodFwiLFxuXCJmb250VmFyaWFudFwiLFxuXCJmb250VmFyaWFudEFsdGVybmF0ZXNcIixcblwiZm9udFZhcmlhbnRDYXBzXCIsXG5cImZvbnRWYXJpYW50RWFzdEFzaWFuXCIsXG5cImZvbnRWYXJpYW50TGlnYXR1cmVzXCIsXG5cImZvbnRWYXJpYW50TnVtZXJpY1wiLFxuXCJmb250VmFyaWFudFBvc2l0aW9uXCIsXG5cImZvbnRWYXJpYXRpb25TZXR0aW5nc1wiLFxuXCJmb250V2VpZ2h0XCIsXG5cImZvcmNlZENvbG9yQWRqdXN0XCIsXG5cImdhcFwiLFxuXCJncmlkXCIsXG5cImdyaWRBcmVhXCIsXG5cImdyaWRBdXRvQ29sdW1uc1wiLFxuXCJncmlkQXV0b0Zsb3dcIixcblwiZ3JpZEF1dG9Sb3dzXCIsXG5cImdyaWRDb2x1bW5cIixcblwiZ3JpZENvbHVtbkVuZFwiLFxuXCJncmlkQ29sdW1uR2FwXCIsXG5cImdyaWRDb2x1bW5TdGFydFwiLFxuXCJncmlkR2FwXCIsXG5cImdyaWRSb3dcIixcblwiZ3JpZFJvd0VuZFwiLFxuXCJncmlkUm93R2FwXCIsXG5cImdyaWRSb3dTdGFydFwiLFxuXCJncmlkVGVtcGxhdGVcIixcblwiZ3JpZFRlbXBsYXRlQXJlYXNcIixcblwiZ3JpZFRlbXBsYXRlQ29sdW1uc1wiLFxuXCJncmlkVGVtcGxhdGVSb3dzXCIsXG5cImhlaWdodFwiLFxuXCJoeXBoZW5hdGVDaGFyYWN0ZXJcIixcblwiaHlwaGVuYXRlTGltaXRDaGFyc1wiLFxuXCJoeXBoZW5zXCIsXG5cImltYWdlT3JpZW50YXRpb25cIixcblwiaW1hZ2VSZW5kZXJpbmdcIixcblwiaW5saW5lU2l6ZVwiLFxuXCJpbnNldFwiLFxuXCJpbnNldEJsb2NrXCIsXG5cImluc2V0QmxvY2tFbmRcIixcblwiaW5zZXRCbG9ja1N0YXJ0XCIsXG5cImluc2V0SW5saW5lXCIsXG5cImluc2V0SW5saW5lRW5kXCIsXG5cImluc2V0SW5saW5lU3RhcnRcIixcblwiaXNvbGF0aW9uXCIsXG5cImp1c3RpZnlDb250ZW50XCIsXG5cImp1c3RpZnlJdGVtc1wiLFxuXCJqdXN0aWZ5U2VsZlwiLFxuXCJsZWZ0XCIsXG5cInJlYWRvbmx5IGxlbmd0aDogbnVtYmVyO1wiLFxuXCJsZXR0ZXJTcGFjaW5nXCIsXG5cImxpZ2h0aW5nQ29sb3JcIixcblwibGluZUJyZWFrXCIsXG5cImxpbmVIZWlnaHRcIixcblwibGlzdFN0eWxlXCIsXG5cImxpc3RTdHlsZUltYWdlXCIsXG5cImxpc3RTdHlsZVBvc2l0aW9uXCIsXG5cImxpc3RTdHlsZVR5cGVcIixcblwibWFyZ2luXCIsXG5cIm1hcmdpbkJsb2NrXCIsXG5cIm1hcmdpbkJsb2NrRW5kXCIsXG5cIm1hcmdpbkJsb2NrU3RhcnRcIixcblwibWFyZ2luQm90dG9tXCIsXG5cIm1hcmdpbklubGluZVwiLFxuXCJtYXJnaW5JbmxpbmVFbmRcIixcblwibWFyZ2luSW5saW5lU3RhcnRcIixcblwibWFyZ2luTGVmdFwiLFxuXCJtYXJnaW5SaWdodFwiLFxuXCJtYXJnaW5Ub3BcIixcblwibWFya2VyXCIsXG5cIm1hcmtlckVuZFwiLFxuXCJtYXJrZXJNaWRcIixcblwibWFya2VyU3RhcnRcIixcblwibWFza1wiLFxuXCJtYXNrQ2xpcFwiLFxuXCJtYXNrQ29tcG9zaXRlXCIsXG5cIm1hc2tJbWFnZVwiLFxuXCJtYXNrTW9kZVwiLFxuXCJtYXNrT3JpZ2luXCIsXG5cIm1hc2tQb3NpdGlvblwiLFxuXCJtYXNrUmVwZWF0XCIsXG5cIm1hc2tTaXplXCIsXG5cIm1hc2tUeXBlXCIsXG5cIm1hdGhEZXB0aFwiLFxuXCJtYXRoU3R5bGVcIixcblwibWF4QmxvY2tTaXplXCIsXG5cIm1heEhlaWdodFwiLFxuXCJtYXhJbmxpbmVTaXplXCIsXG5cIm1heFdpZHRoXCIsXG5cIm1pbkJsb2NrU2l6ZVwiLFxuXCJtaW5IZWlnaHRcIixcblwibWluSW5saW5lU2l6ZVwiLFxuXCJtaW5XaWR0aFwiLFxuXCJtaXhCbGVuZE1vZGVcIixcblwib2JqZWN0Rml0XCIsXG5cIm9iamVjdFBvc2l0aW9uXCIsXG5cIm9mZnNldFwiLFxuXCJvZmZzZXRBbmNob3JcIixcblwib2Zmc2V0RGlzdGFuY2VcIixcblwib2Zmc2V0UGF0aFwiLFxuXCJvZmZzZXRQb3NpdGlvblwiLFxuXCJvZmZzZXRSb3RhdGVcIixcblwib3BhY2l0eVwiLFxuXCJvcmRlclwiLFxuXCJvcnBoYW5zXCIsXG5cIm91dGxpbmVcIixcblwib3V0bGluZUNvbG9yXCIsXG5cIm91dGxpbmVPZmZzZXRcIixcblwib3V0bGluZVN0eWxlXCIsXG5cIm91dGxpbmVXaWR0aFwiLFxuXCJvdmVyZmxvd1wiLFxuXCJvdmVyZmxvd0FuY2hvclwiLFxuXCJvdmVyZmxvd0Jsb2NrXCIsXG5cIm92ZXJmbG93Q2xpcE1hcmdpblwiLFxuXCJvdmVyZmxvd0lubGluZVwiLFxuXCJvdmVyZmxvd1dyYXBcIixcblwib3ZlcmZsb3dYXCIsXG5cIm92ZXJmbG93WVwiLFxuXCJvdmVyc2Nyb2xsQmVoYXZpb3JcIixcblwib3ZlcnNjcm9sbEJlaGF2aW9yQmxvY2tcIixcblwib3ZlcnNjcm9sbEJlaGF2aW9ySW5saW5lXCIsXG5cIm92ZXJzY3JvbGxCZWhhdmlvclhcIixcblwib3ZlcnNjcm9sbEJlaGF2aW9yWVwiLFxuXCJwYWRkaW5nXCIsXG5cInBhZGRpbmdCbG9ja1wiLFxuXCJwYWRkaW5nQmxvY2tFbmRcIixcblwicGFkZGluZ0Jsb2NrU3RhcnRcIixcblwicGFkZGluZ0JvdHRvbVwiLFxuXCJwYWRkaW5nSW5saW5lXCIsXG5cInBhZGRpbmdJbmxpbmVFbmRcIixcblwicGFkZGluZ0lubGluZVN0YXJ0XCIsXG5cInBhZGRpbmdMZWZ0XCIsXG5cInBhZGRpbmdSaWdodFwiLFxuXCJwYWRkaW5nVG9wXCIsXG5cInBhZ2VcIixcblwicGFnZUJyZWFrQWZ0ZXJcIixcblwicGFnZUJyZWFrQmVmb3JlXCIsXG5cInBhZ2VCcmVha0luc2lkZVwiLFxuXCJwYWludE9yZGVyXCIsXG5cInJlYWRvbmx5IHBhcmVudFJ1bGU6IENTU1J1bGUgfCBudWxsO1wiLFxuXCJwZXJzcGVjdGl2ZVwiLFxuXCJwZXJzcGVjdGl2ZU9yaWdpblwiLFxuXCJwbGFjZUNvbnRlbnRcIixcblwicGxhY2VJdGVtc1wiLFxuXCJwbGFjZVNlbGZcIixcblwicG9pbnRlckV2ZW50c1wiLFxuXCJwb3NpdGlvblwiLFxuXCJwcmludENvbG9yQWRqdXN0XCIsXG5cInF1b3Rlc1wiLFxuXCJyXCIsXG5cInJlc2l6ZVwiLFxuXCJyaWdodFwiLFxuXCJyb3RhdGVcIixcblwicm93R2FwXCIsXG5cInJ1YnlBbGlnblwiLFxuXCJydWJ5UG9zaXRpb25cIixcblwicnhcIixcblwicnlcIixcblwic2NhbGVcIixcblwic2Nyb2xsQmVoYXZpb3JcIixcblwic2Nyb2xsTWFyZ2luXCIsXG5cInNjcm9sbE1hcmdpbkJsb2NrXCIsXG5cInNjcm9sbE1hcmdpbkJsb2NrRW5kXCIsXG5cInNjcm9sbE1hcmdpbkJsb2NrU3RhcnRcIixcblwic2Nyb2xsTWFyZ2luQm90dG9tXCIsXG5cInNjcm9sbE1hcmdpbklubGluZVwiLFxuXCJzY3JvbGxNYXJnaW5JbmxpbmVFbmRcIixcblwic2Nyb2xsTWFyZ2luSW5saW5lU3RhcnRcIixcblwic2Nyb2xsTWFyZ2luTGVmdFwiLFxuXCJzY3JvbGxNYXJnaW5SaWdodFwiLFxuXCJzY3JvbGxNYXJnaW5Ub3BcIixcblwic2Nyb2xsUGFkZGluZ1wiLFxuXCJzY3JvbGxQYWRkaW5nQmxvY2tcIixcblwic2Nyb2xsUGFkZGluZ0Jsb2NrRW5kXCIsXG5cInNjcm9sbFBhZGRpbmdCbG9ja1N0YXJ0XCIsXG5cInNjcm9sbFBhZGRpbmdCb3R0b21cIixcblwic2Nyb2xsUGFkZGluZ0lubGluZVwiLFxuXCJzY3JvbGxQYWRkaW5nSW5saW5lRW5kXCIsXG5cInNjcm9sbFBhZGRpbmdJbmxpbmVTdGFydFwiLFxuXCJzY3JvbGxQYWRkaW5nTGVmdFwiLFxuXCJzY3JvbGxQYWRkaW5nUmlnaHRcIixcblwic2Nyb2xsUGFkZGluZ1RvcFwiLFxuXCJzY3JvbGxTbmFwQWxpZ25cIixcblwic2Nyb2xsU25hcFN0b3BcIixcblwic2Nyb2xsU25hcFR5cGVcIixcblwic2Nyb2xsYmFyQ29sb3JcIixcblwic2Nyb2xsYmFyR3V0dGVyXCIsXG5cInNjcm9sbGJhcldpZHRoXCIsXG5cInNoYXBlSW1hZ2VUaHJlc2hvbGRcIixcblwic2hhcGVNYXJnaW5cIixcblwic2hhcGVPdXRzaWRlXCIsXG5cInNoYXBlUmVuZGVyaW5nXCIsXG5cInN0b3BDb2xvclwiLFxuXCJzdG9wT3BhY2l0eVwiLFxuXCJzdHJva2VcIixcblwic3Ryb2tlRGFzaGFycmF5XCIsXG5cInN0cm9rZURhc2hvZmZzZXRcIixcblwic3Ryb2tlTGluZWNhcFwiLFxuXCJzdHJva2VMaW5lam9pblwiLFxuXCJzdHJva2VNaXRlcmxpbWl0XCIsXG5cInN0cm9rZU9wYWNpdHlcIixcblwic3Ryb2tlV2lkdGhcIixcblwidGFiU2l6ZVwiLFxuXCJ0YWJsZUxheW91dFwiLFxuXCJ0ZXh0QWxpZ25cIixcblwidGV4dEFsaWduTGFzdFwiLFxuXCJ0ZXh0QW5jaG9yXCIsXG5cInRleHRCb3hcIixcblwidGV4dEJveEVkZ2VcIixcblwidGV4dEJveFRyaW1cIixcblwidGV4dENvbWJpbmVVcHJpZ2h0XCIsXG5cInRleHREZWNvcmF0aW9uXCIsXG5cInRleHREZWNvcmF0aW9uQ29sb3JcIixcblwidGV4dERlY29yYXRpb25MaW5lXCIsXG5cInRleHREZWNvcmF0aW9uU2tpcElua1wiLFxuXCJ0ZXh0RGVjb3JhdGlvblN0eWxlXCIsXG5cInRleHREZWNvcmF0aW9uVGhpY2tuZXNzXCIsXG5cInRleHRFbXBoYXNpc1wiLFxuXCJ0ZXh0RW1waGFzaXNDb2xvclwiLFxuXCJ0ZXh0RW1waGFzaXNQb3NpdGlvblwiLFxuXCJ0ZXh0RW1waGFzaXNTdHlsZVwiLFxuXCJ0ZXh0SW5kZW50XCIsXG5cInRleHRPcmllbnRhdGlvblwiLFxuXCJ0ZXh0T3ZlcmZsb3dcIixcblwidGV4dFJlbmRlcmluZ1wiLFxuXCJ0ZXh0U2hhZG93XCIsXG5cInRleHRUcmFuc2Zvcm1cIixcblwidGV4dFVuZGVybGluZU9mZnNldFwiLFxuXCJ0ZXh0VW5kZXJsaW5lUG9zaXRpb25cIixcblwidGV4dFdyYXBcIixcblwidGV4dFdyYXBNb2RlXCIsXG5cInRleHRXcmFwU3R5bGVcIixcblwidG9wXCIsXG5cInRvdWNoQWN0aW9uXCIsXG5cInRyYW5zZm9ybVwiLFxuXCJ0cmFuc2Zvcm1Cb3hcIixcblwidHJhbnNmb3JtT3JpZ2luXCIsXG5cInRyYW5zZm9ybVN0eWxlXCIsXG5cInRyYW5zaXRpb25cIixcblwidHJhbnNpdGlvbkJlaGF2aW9yXCIsXG5cInRyYW5zaXRpb25EZWxheVwiLFxuXCJ0cmFuc2l0aW9uRHVyYXRpb25cIixcblwidHJhbnNpdGlvblByb3BlcnR5XCIsXG5cInRyYW5zaXRpb25UaW1pbmdGdW5jdGlvblwiLFxuXCJ0cmFuc2xhdGVcIixcblwidW5pY29kZUJpZGlcIixcblwidXNlclNlbGVjdFwiLFxuXCJ2ZWN0b3JFZmZlY3RcIixcblwidmVydGljYWxBbGlnblwiLFxuXCJ2aWV3VHJhbnNpdGlvbkNsYXNzXCIsXG5cInZpZXdUcmFuc2l0aW9uTmFtZVwiLFxuXCJ2aXNpYmlsaXR5XCIsXG5cIndlYmtpdEFsaWduQ29udGVudFwiLFxuXCJ3ZWJraXRBbGlnbkl0ZW1zXCIsXG5cIndlYmtpdEFsaWduU2VsZlwiLFxuXCJ3ZWJraXRBbmltYXRpb25cIixcblwid2Via2l0QW5pbWF0aW9uRGVsYXlcIixcblwid2Via2l0QW5pbWF0aW9uRGlyZWN0aW9uXCIsXG5cIndlYmtpdEFuaW1hdGlvbkR1cmF0aW9uXCIsXG5cIndlYmtpdEFuaW1hdGlvbkZpbGxNb2RlXCIsXG5cIndlYmtpdEFuaW1hdGlvbkl0ZXJhdGlvbkNvdW50XCIsXG5cIndlYmtpdEFuaW1hdGlvbk5hbWVcIixcblwid2Via2l0QW5pbWF0aW9uUGxheVN0YXRlXCIsXG5cIndlYmtpdEFuaW1hdGlvblRpbWluZ0Z1bmN0aW9uXCIsXG5cIndlYmtpdEFwcGVhcmFuY2VcIixcblwid2Via2l0QmFja2ZhY2VWaXNpYmlsaXR5XCIsXG5cIndlYmtpdEJhY2tncm91bmRDbGlwXCIsXG5cIndlYmtpdEJhY2tncm91bmRPcmlnaW5cIixcblwid2Via2l0QmFja2dyb3VuZFNpemVcIixcblwid2Via2l0Qm9yZGVyQm90dG9tTGVmdFJhZGl1c1wiLFxuXCJ3ZWJraXRCb3JkZXJCb3R0b21SaWdodFJhZGl1c1wiLFxuXCJ3ZWJraXRCb3JkZXJSYWRpdXNcIixcblwid2Via2l0Qm9yZGVyVG9wTGVmdFJhZGl1c1wiLFxuXCJ3ZWJraXRCb3JkZXJUb3BSaWdodFJhZGl1c1wiLFxuXCJ3ZWJraXRCb3hBbGlnblwiLFxuXCJ3ZWJraXRCb3hGbGV4XCIsXG5cIndlYmtpdEJveE9yZGluYWxHcm91cFwiLFxuXCJ3ZWJraXRCb3hPcmllbnRcIixcblwid2Via2l0Qm94UGFja1wiLFxuXCJ3ZWJraXRCb3hTaGFkb3dcIixcblwid2Via2l0Qm94U2l6aW5nXCIsXG5cIndlYmtpdEZpbHRlclwiLFxuXCJ3ZWJraXRGbGV4XCIsXG5cIndlYmtpdEZsZXhCYXNpc1wiLFxuXCJ3ZWJraXRGbGV4RGlyZWN0aW9uXCIsXG5cIndlYmtpdEZsZXhGbG93XCIsXG5cIndlYmtpdEZsZXhHcm93XCIsXG5cIndlYmtpdEZsZXhTaHJpbmtcIixcblwid2Via2l0RmxleFdyYXBcIixcblwid2Via2l0SnVzdGlmeUNvbnRlbnRcIixcblwid2Via2l0TGluZUNsYW1wXCIsXG5cIndlYmtpdE1hc2tcIixcblwid2Via2l0TWFza0JveEltYWdlXCIsXG5cIndlYmtpdE1hc2tCb3hJbWFnZU91dHNldFwiLFxuXCJ3ZWJraXRNYXNrQm94SW1hZ2VSZXBlYXRcIixcblwid2Via2l0TWFza0JveEltYWdlU2xpY2VcIixcblwid2Via2l0TWFza0JveEltYWdlU291cmNlXCIsXG5cIndlYmtpdE1hc2tCb3hJbWFnZVdpZHRoXCIsXG5cIndlYmtpdE1hc2tDbGlwXCIsXG5cIndlYmtpdE1hc2tDb21wb3NpdGVcIixcblwid2Via2l0TWFza0ltYWdlXCIsXG5cIndlYmtpdE1hc2tPcmlnaW5cIixcblwid2Via2l0TWFza1Bvc2l0aW9uXCIsXG5cIndlYmtpdE1hc2tSZXBlYXRcIixcblwid2Via2l0TWFza1NpemVcIixcblwid2Via2l0T3JkZXJcIixcblwid2Via2l0UGVyc3BlY3RpdmVcIixcblwid2Via2l0UGVyc3BlY3RpdmVPcmlnaW5cIixcblwid2Via2l0VGV4dEZpbGxDb2xvclwiLFxuXCJ3ZWJraXRUZXh0U2l6ZUFkanVzdFwiLFxuXCJ3ZWJraXRUZXh0U3Ryb2tlXCIsXG5cIndlYmtpdFRleHRTdHJva2VDb2xvclwiLFxuXCJ3ZWJraXRUZXh0U3Ryb2tlV2lkdGhcIixcblwid2Via2l0VHJhbnNmb3JtXCIsXG5cIndlYmtpdFRyYW5zZm9ybU9yaWdpblwiLFxuXCJ3ZWJraXRUcmFuc2Zvcm1TdHlsZVwiLFxuXCJ3ZWJraXRUcmFuc2l0aW9uXCIsXG5cIndlYmtpdFRyYW5zaXRpb25EZWxheVwiLFxuXCJ3ZWJraXRUcmFuc2l0aW9uRHVyYXRpb25cIixcblwid2Via2l0VHJhbnNpdGlvblByb3BlcnR5XCIsXG5cIndlYmtpdFRyYW5zaXRpb25UaW1pbmdGdW5jdGlvblwiLFxuXCJ3ZWJraXRVc2VyU2VsZWN0XCIsXG5cIndoaXRlU3BhY2VcIixcblwid2hpdGVTcGFjZUNvbGxhcHNlXCIsXG5cIndpZG93c1wiLFxuXCJ3aWR0aFwiLFxuXCJ3aWxsQ2hhbmdlXCIsXG5cIndvcmRCcmVha1wiLFxuXCJ3b3JkU3BhY2luZ1wiLFxuXCJ3b3JkV3JhcFwiLFxuXCJ3cml0aW5nTW9kZVwiLFxuXCJ4XCIsXG5cInlcIixcblwiekluZGV4XCIsXG5cInpvb21cIl07XG5cbmNvbnN0IGtleXMyID0gW107XG5mb3IgKGNvbnN0IGtleSBvZiBrZXlzKSB7XG4gIGZvciAobGV0IGkgPSAwOyBpIDwgMTAwOyBpKyspIHtcbiAgICBrZXlzMi5wdXNoKGtleSk7XG4gIH1cbn1cblxucmV0dXJuIGtleXMyOyJ9fQ) in Firefox and Chrome. In Chrome has/get is consistently 10 to 20% faster. In Firefox it's consistently 40 to 50% faster.

trueadm (approved):
Great stuff

packages/ripple/src/utils/normalize_css_property_name.js
leonidaz: Just for posterity, you would not call set each time as you would store the result, compare to undefined and do an early return with the stored get value. 

But looks like has is very optimized anyway.

Yusyuriv: Oh, you're right, I didn't think of that. Your method gets an additional 10 to 15% improvement in both Firefox and Chrome. Should I create a new PR?

trueadm: Sure

leonidaz:
@Yusyuriv  what’s your discord handle?

Yusyuriv:
@leonidaz elvenprincess

PR #363, opened 2025-09-28
Fix portal cleanup

anubra266:
This PR stores a reference to portaled nodes, and destroys them when portal is cleaned up.

### DIFF
--- a/packages/ripple/src/runtime/internal/client/portal.js
+++ b/packages/ripple/src/runtime/internal/client/portal.js
@@ -1,6 +1,6 @@
 /** @import { Block } from '#client' */
 
-import { branch, destroy_block, render } from './blocks.js';
+import { branch, destroy_block, remove_block_dom, render } from './blocks.js';
 import { UNINITIALIZED } from './constants.js';
 import { handle_root_events } from './events.js';
 import { create_text } from './operations.js';
@@ -19,6 +19,10 @@ export function Portal(_, props) {
   var b = null;
   /** @type {Text | null} */
   var anchor = null;
+  /** @type {Node | null} */
+  var domStart = null;
+  /** @type {Node | null} */
+  var domEnd = null;
 
   render(() => {
     if (target === (target = props.target)) return;
@@ -28,6 +32,8 @@ export function Portal(_, props) {
       destroy_block(b);
     }
 
+    domStart = domEnd = null;
+
     anchor = create_text();
     /** @type {Element} */ (target).append(anchor);
 
@@ -35,9 +41,17 @@ export function Portal(_, props) {
 
     b = branch(() => /** @type {(anchor: Node) => void} */ (children)(/** @type {Text} */ (anchor)));
 
+    if (b && b.s) {
+      domStart = b.s.start;
+      domEnd = b.s.end;
+    }
+
     return () => {
       cleanup_events();
       /** @type {Text} */ (anchor).remove();
+      if(domStart !== null && domEnd !== null) {
+        remove_block_dom(domStart, domEnd);
+      }
     };
   });
 }
--- a/packages/ripple/src/runtime/internal/client/runtime.js
+++ b/packages/ripple/src/runtime/internal/client/runtime.js
@@ -752,7 +752,7 @@ export function set(tracked, value, block) {
 	if (value !== old_value) {
 		var tracked_block = tracked.b;
 
-		if ((block.f & CONTAINS_TEARDOWN) !== 0) {
+		if (block && (block.f & CONTAINS_TEARDOWN) !== 0) {
 			if (teardown) {
 				old_values.set(tracked, value);
 			} else {


leonidaz:
nice! we'd definitely want at least a couple of tests though for any future regressions.

anubra266:
Done @leonidaz

packages/ripple/src/runtime/internal/client/portal.js
```diff
@@ -28,16 +32,26 @@ export function Portal(_, props) {
...
     }
 
+    domStart = domEnd = null;
+
     anchor = create_text();
     /** @type {Element} */ (target).append(anchor);
 
     const cleanup_events = handle_root_events(/** @type {Element} */ (target));
 
     b = branch(() => /** @type {(anchor: Node) => void} */ (children)(/** @type {Text} */ (anchor)));
```
leonidaz: unrelated but these inline types here make it impossible to read the code, we should just `// @ts-ignore` it.

packages/ripple/tests/client/portal.test.ripple
```diff
@@ -0,0 +1,208 @@
...
+
+		expect(portalElement.textContent).toContain('Count: 1');
+	});
+
+	// Cleanup any leftover portal content after each test
+	afterEach(() => {
+		// Remove any portal content that might be left in document.body
+		const portals = document.body.querySelectorAll('.portal-content, .body-portal, .conditional-portal, .toggle-portal, .moving-portal, .portal1, .portal2, .reactive-portal');
+		portals.forEach(el => el.remove());
+	});
```
leonidaz: can't we just use the same class in tests and move this section to the top, combining it with the other one there?

packages/ripple/src/runtime/internal/client/portal.js
```diff
@@ -28,16 +32,26 @@ export function Portal(_, props) {
...
 
+    if (b && b.s) {
+      domStart = b.s.start;
+      domEnd = b.s.end;
+    }
```
leonidaz:
```suggestion
    domStart = b?.s?.start;
    domEnd = b?.s?.end;
```

packages/ripple/src/runtime/internal/client/portal.js
```diff
@@ -28,16 +32,26 @@ export function Portal(_, props) {
...
+
     return () => {
       cleanup_events();
       /** @type {Text} */ (anchor).remove();
+      if(domStart !== null && domEnd !== null) {
```
leonidaz: seems like these can also be undefined.


```suggestion
      if(domStart != null && domEnd != null) {
```

packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -752,7 +752,7 @@ export function set(tracked, value, block) {
 	if (value !== old_value) {
 		var tracked_block = tracked.b;
 
-		if ((block.f & CONTAINS_TEARDOWN) !== 0) {
+		if (block && (block.f & CONTAINS_TEARDOWN) !== 0) {
```
leonidaz: is this possible here that the block doesn't exist?

packages/ripple/tests/client/portal.test.ripple
```diff
@@ -0,0 +1,208 @@
...
+		expect(() => {
+			document.body.querySelector('button').click();
+			flushSync();
+		}).not.toThrow();
+
+		// Portal content should be cleaned up
+		expect(document.body.querySelector('.toggle-portal')).toBeNull();
+	});
+
+	it.skip('updates portal content when target changes', () => {
```
leonidaz: do we want to keep this or remove this test?

packages/ripple/src/runtime/internal/client/runtime.js
anubra266: Yeah @leonidaz in fact `cannot access f of undefined` was the inittial error. 😁

leonidaz: right, was just thinking if we're doing something wrong to cause this...

trueadm: Prefer `block !== null` here

anubra266: will need to look into this more @trueadm 
CleanShot 2025-09-29 at 01 31 25@2x

It's happening here:
https://github.com/anubra266/ripple/blob/fix-portal-cleanup/packages/ripple/src/runtime/internal/client/portal.js#L44

trueadm: Actually hold on, `block` should always be set here. This points to a bug where we're not providing `block`. Do you have a full working example, as this doesn't seem confined to this PR

anubra266: Right...it was happening before this PR, but I could only reproduce it in the portal cleanup. Seems I don't understand the destroy logic of the blocks as much as I thought

trueadm: No i mean the problem here isn't this line. If you go to the click event handler, the problem is there It should be passing in the `block` to `set`.

trueadm: Let's remove this change. I'll look into route cause of the issue

packages/ripple/src/runtime/internal/client/portal.js
```diff
@@ -19,6 +19,10 @@ export function Portal(_, props) {
   var b = null;
   /** @type {Text | null} */
   var anchor = null;
+  /** @type {Node | null} */
+  var domStart = null;
```
trueadm: we prefer snake case in the compiler/runtime

anubra266:
```suggestion
  var dom_start = null;
```

packages/ripple/src/runtime/internal/client/portal.js
```diff
@@ -19,6 +19,10 @@ export function Portal(_, props) {
...
+  /** @type {Node | null} */
+  var domEnd = null;
```
anubra266:
```suggestion
  var dom_end = null;

File: packages/ripple/src/runtime/internal/client/portal.js [hunk 9]
```diff
@@ -28,16 +32,29 @@ export function Portal(_, props) {
       destroy_block(b);
     }
 
+    if (anchor !== null) {
+      anchor.remove();
+    }
+
+    domStart = domEnd = null;
```
anubra266:
```suggestion
    dom_start = dom_end = null;

File: packages/ripple/src/runtime/internal/client/portal.js [hunk 10]
```diff
@@ -28,16 +32,29 @@ export function Portal(_, props) {
...
     /** @type {Element} */ (target).append(anchor);
 
     const cleanup_events = handle_root_events(/** @type {Element} */ (target));
 
-    b = branch(() => /** @type {(anchor: Node) => void} */ (children)(/** @type {Text} */ (anchor)));
+    // @ts-ignore
+    b = branch(() =>  children(anchor));
+
+    domStart = b?.s?.start;
+    domEnd = b?.s?.end;
```
anubra266:
```suggestion
    dom_end = b?.s?.end;

File: packages/ripple/src/runtime/internal/client/portal.js [hunk 11]
```diff
@@ -28,16 +32,29 @@ export function Portal(_, props) {
...
     anchor = create_text();
     /** @type {Element} */ (target).append(anchor);
 
     const cleanup_events = handle_root_events(/** @type {Element} */ (target));
 
-    b = branch(() => /** @type {(anchor: Node) => void} */ (children)(/** @type {Text} */ (anchor)));
+    // @ts-ignore
+    b = branch(() =>  children(anchor));
+
+    domStart = b?.s?.start;
```
anubra266:
```suggestion
    dom_start = b?.s?.start;

File: packages/ripple/src/runtime/internal/client/portal.js [hunk 12]
```diff
@@ -28,16 +32,29 @@ export function Portal(_, props) {
...
+    domEnd = b?.s?.end;
 
     return () => {
       cleanup_events();
       /** @type {Text} */ (anchor).remove();
+      if(domStart && domEnd) {
```
anubra266:
```suggestion
      if(dom_start && dom_end) {

File: packages/ripple/src/runtime/internal/client/portal.js [hunk 13]
```diff
@@ -28,16 +32,29 @@ export function Portal(_, props) {
...
+        remove_block_dom(domStart, domEnd);
```
anubra266:
```suggestion
        remove_block_dom(dom_start, dom_end);

File: packages/ripple/src/runtime/internal/client/portal.js [hunk 14]
```diff
@@ -28,16 +32,29 @@ export function Portal(_, props) {
...
+    dom_start = dom_end = null;
+
     anchor = create_text();
     /** @type {Element} */ (target).append(anchor);
 
     const cleanup_events = handle_root_events(/** @type {Element} */ (target));
 
-    b = branch(() => /** @type {(anchor: Node) => void} */ (children)(/** @type {Text} */ (anchor)));
+    // @ts-ignore
+    b = branch(() =>  children(anchor));
```
trueadm: I know the issue now, this is missing `props` and `block`, here `props` is `{}` and `block` is `active_block` (you'll need to cast it to `Block`) as we know its active here

It was a bug in the original implementation too

trueadm:
I pushed a fix

leonidaz:
Let's merge it then, seems ready to go.

Issue #361, opened 2025-09-28
[Feature]: raw html rendering supports SVG/MathML


leonidaz:
### Feature Category

Performance Enhancement

### Problem Statement

This was left as a TODO

### Proposed Solution

`{html str}` properly renders svg and mathml stringified elements

### Checklist

- [x] I've searched for existing feature requests
- [ ] I'm willing to help implement this feature

PR #371, opened 2025-09-29
adds support for {html str} inside svg and mathml

leonidaz:
- ignore white space changes when looking at the diff

implements #361


packages/ripple/src/runtime/internal/client/html.js
```diff
@@ -8,14 +8,14 @@ import { assign_nodes, create_fragment_from_html } from './template.js';
 /**
  * Renders dynamic HTML content into the DOM by inserting it before the anchor node.
  * Manages the lifecycle of HTML blocks, removing old content and inserting new content.
- * 
+ *
  * TODO handle SVG/MathML
```
trueadm: This can go now

PR #373, opened 2025-09-29
test: Add test for mutating tracked value inside computed track() evaluation

Yusyuriv:
A recent commit disallowed mutating tracked values inside computed `track()`. This test tries to mutate a tracked value inside a computed `track()`, and verifies that the error gets thrown.

Side note: is there a better way to catch errors and get their text in the test? This is the only solution I came up with that seemed to work.

### DIFF
--- a/packages/ripple/tests/client/basic.test.ripple
+++ b/packages/ripple/tests/client/basic.test.ripple
@@ -1760,5 +1760,46 @@ describe('basic client', () => {
 		await promise; 
 		expect(p.textContent).toBe('1');
 	});
+
+	it('errors on mutating tracked value inside computed track() evaluation', () => {
+		component Basic() {
+			let count = track(0);
+
+			let p = track();
+			function pRef(node) {
+				@p = node;
+
+				return () => {
+					@p = null;
+				};
+			}
+
+			const doubled = track(() => {
+				try {
+					@count *= 2;
+				} catch (e) {
+					if(@p) {
+						@p.textContent = e.message;
+					}
+				}
+			});
+
+			<p class='error' {ref pRef} />
+
+			<p>{@doubled}</p>
+
+			<button onClick={() => { @count++; }}>{@count}</button>
+		}
+
+		render(Basic);
+
+		const button = container.querySelector('button');
+		const error = container.querySelector('.error');
+
+		button.click();
+		flushSync();
+
+		expect(error.textContent).toContain('Assignments or updates to tracked values are not allowed during computed "track(() => ...)" evaluation');
+	});
 });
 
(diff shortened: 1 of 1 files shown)


trueadm:
That's one way, another is to use the Vitest `expect` `toThrowError` API

Yusyuriv:
> That's one way, another is to use the Vitest `expect` `toThrowError` API

That would be the obvious choice, so that's what I tried first, but the test was failing:
```ts
expect(() => {
	button.click();
	flushSync();
}).toThrowError();
```
```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

Vitest caught 1 unhandled error during the test run.                       
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.                                             

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: Assignments or updates to tracked values are not allowed during computed "track(() => ...)" evaluation    
```

trueadm:
Yeah lets just do this then. You can add an event listener for unhandled errors because of the promise, but it's likely more work.

Yusyuriv:
Okay, just thought I'd ask because you might have already encountered this issue somewhere else. I think it's an okay solution for a one-off test. If we need to test more errors in the future, I can look into simplifying that.

One last question: how does PR merge process work? Do I merge it myself after someone reviews it, or do I have to wait for someone else to merge it?

trueadm:
Generally speaking we wait for a review from someone else on the team. Unless it's trivial, like docs, or a small fix in which case you can merge yourself. Ask on Discord for a review if needed and assign reviewers me and @leonidaz if in doubt

leonidaz:
it's probably good to add a test for doing the same thing inside track's get / set.

Btw, if someone does mutate inside computed, can they do so inside of `untrack`?  If so, another set of tests would be great..

trueadm (approved):

Yusyuriv:
I went the path of least resistance and just created a global `error` variable. Components can set it, and it gets cleared before each test. Seems to work well, and doesn't require a lot of boilerplate in each test component.

Is allowing changes to tracked variables inside `track()` setters the correct behavior? I wrote that test to check that the error is not thrown because that's the current behavior. Is this intended?

leonidaz:
> allowing changes to tracked variables inside `track()` setters

I think so, we discussed with Dominic the other day that if would be very useful for creating light weight models.

PR #383, opened 2025-09-30
edit playground actions styles

hatemhosny:
This PR improves styles for action buttons above the playground to be more consistent with the website styles.


leonidaz (changes requested):
I think with this change we want the gray line back in the header, actually it should've been there from the beginning.

Image

There is a div.divider which has div.divider-line inside of it, it looks like the styling is being overriden on this page.  

Once enabled, the contents of the playground top bar with examples, etc. should also be vertically centered.

hatemhosny:
Thank you @leonidaz 
Suggested changes were applied.

PR #386, opened 2025-09-30
change createContext to new Context

leonidaz:
- changes syntax for `createContext()` to use `new Context()`.
- clarity that it's not just a global context that is being created and multiple ones can exist
- move away from the React association with useContext, creating provider, etc.


trueadm:
Did you check SSR context too?

leonidaz:
yeah, I changed the api for the server as well.

PR #389, opened 2025-10-01
playground styles

hatemhosny:
This is a follow up PR for #383 

Just small style improvements:
- It removes the main playground border to avoid having double borders (with header) while loading
- better vertical alignment of playground buttons


leonidaz (approved):
👍

PR #390, opened 2025-10-01
Migrate to the TypeScript plugin

johnsoncodehk:


.vscode/launch.json
```diff
@@ -9,6 +9,7 @@
 			"name": "Extension",
 			"type": "extensionHost",
 			"request": "launch",
+			"autoAttachChildProcesses": true,
```
johnsoncodehk, line 12: This will enable breakpoints to take effect in `typescript-plugin-ripple`.

PR #391, opened 2025-10-01
Added ripple icon entry in contributes schema for both light & dark mode

Atabic:
Hi, 
I saw while playing around with ripple that no logo is shown for .ripple files. So I thought of adding this for both light and dark mode. 
Maybe we can later decided separate icons for both modes. 

I'm not sure whether I should package the extension again or not and update that too with in this PR. 

It will be a huge help if someone can guide a little on what else I should do to get this PR merged.

P.S: This is my first PR to any library :)

Screenshot 2025-10-02 at 4 55 01 AM

### DIFF
--- a/packages/ripple-vscode-plugin/package.json
+++ b/packages/ripple-vscode-plugin/package.json
@@ -41,6 +41,10 @@
         "extensions": [
           ".ripple"
         ],
+        "icon": {
+          "light": "icons/logo.png",
+          "dark": "icons/logo.png"
+        },
         "configuration": "./language-configuration.json"
       }
     ],


trueadm (approved):
Thank you

JakeIsMeh:
I know this has been merged already, but for this small of a context we should be using the mini variant of the icon, same as the favicon on the site right now, rather than the big version.

leonidaz:
Yeah, we should for sure use the other icon, below, the current one doesn't look good in the file explorer: way too busy.

Here's the svg for it, if we agree, let's swap them:

favicon

trueadm:
Agreed, let's create another PR follow up?

Atabic:
Sure, I'll create another PR for this

PR #395, opened 2025-10-02
explicitly call out "slots" in docs

JakeIsMeh:
also added a redirect for `/docs`

### DIFF
--- a/website/docs/guide/components.md
+++ b/website/docs/guide/components.md
@@ -78,19 +78,22 @@ export component App() {
 }
 ```
 
-### "Slots" Pattern
+## Example: Card Component Using Child Composition
 
-With all that in place, we can now recreate the pattern of "slots" from Vue/Web
-Components, "render props" from React, and "snippets" from Svelte:
+Using what we've learnt, let's make a versatile card component that can display
+an optional header and footer.
+
+This is a common pattern you'll see achieved with "slots" from Vue/Web
+Components, "render props" from React, and "snippets" from Svelte.
 
 ```ripple
 component Card({ children, Header, Footer }) {
 	<fieldset>
-		<Header />  // <-- Slot named `Header`
+		<Header />
 		<hr />
 		<children />
 		<hr />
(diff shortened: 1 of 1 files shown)


leonidaz:
I like the examples, but I personally think we should not talk about slots. Instead I'd want to push the concept that everything is a prop and here's a pattern for this and here's one for that...

trueadm:
I'm not sure we should be calling this the slot pattern – slots have a very strange and odd attachment to web components. Svelte 5 has completely distanced itself from slots too because of the confusion they cause. I think we're best avoiding the terminology completely.

trueadm:
Maybe we want to phrase it as named children composition or something with composition.

leonidaz:
I think it deserves a separate page, `Component Composition`, with various examples and techniques.

JakeIsMeh:
> I'm not sure we should be calling this the slot pattern – slots have a very strange and odd attachment to web components. Svelte 5 has completely distanced itself from slots too because of the confusion they cause. I think we're best avoiding the terminology completely.

I'm bringing up the term "slots" (and "snippets") on purpose, since we've had multiple instances of people asking about slots as a feature when it's already possible. Perhaps I should add "render props" too...


> I think it deserves a separate page, `Component Composition`, with various examples and techniques.

I'm not sure what else we could demonstrate? Ripple's strength is in its composability. Strong, simple examples should be enough to teach concepts that the user can further compose on their own.

trueadm:
We can stipulate in a note that these concepts are similar to render props, snippets and slots in other frameworks. However that can literally be a one line comment that gets indexed in case people search for such terms.

JakeIsMeh:
> We can stipulate in a note that these concepts are similar to render props, snippets and slots in other frameworks. However that can literally be a one line comment that gets indexed in case people search for such terms.

Would you mind suggesting a diff for what you want it to say specifically? There is only a single sentence calling out those other terms, and the example after brings together the aforementioned concepts into a common scenario.

trueadm (approved):
Thank you

PR #396, opened 2025-10-02
add playground title

hatemhosny:
This PR adds playground title.
The title also persists in the shared URL.

It can be used later on social media cards (using open graph tags generated on server-side) for shared URLs.


leonidaz:
Nice! Would be great to display the title on mobile. I’m thinking we can put the version selector on mobile into the settings and put the title in its place.

hatemhosny:
Thank you @leonidaz 
On mobile, I now use version selector as an input in settings menu to avoid nested menus (which also didn't work well with the used UI components).
Please have a look.

leonidaz (approved):
awesome! thank you!

PR #403, opened 2025-10-04
fix playground scroll issue

hatemhosny:
The playground page fills the viewport and has `overflow: hidden` to avoid showing a scrollbar.

However, on screens with small height (e.g. on mobile with landscape orientation, or on desktop with dev tools open), scrolling is allowed to avoid hiding the lower part of the playground.

I found an issue where on mobile with landscape orientation, user scrolls down, then changes orientation to landscape and is not able to scroll back to top. This PR fixes this!


leonidaz (approved):
👍

PR #405, opened 2025-10-04
feat: Add TextMate bundle for Ripple syntax highlighting in editors

Yusyuriv:
WebStorm:
image

Sublime Text:
image


JakeIsMeh:
perhaps we can (temporarily, until we have an org) put this under /assets, since it's not necessarily an npm package. 

it should also be named `ripple.tmbundle` to match textmate convention.

and if you don't mind, document this under `website/docs/quick-start`

trueadm:
I’ve updated the syntax recently FYI

JakeIsMeh (approved):

Issue #266, opened 2025-09-21
[Feature]: Implement Dynamic Elements


tijnjh:
### Feature Category

Syntax/Language Improvement

### Problem Statement

i would like to be able to conditionally change the tag of an element

### Proposed Solution

maybe something like this? just like react

```jsx
component CustomTag(props: { tag: string }) {
  const Tag = props.tag;
  
  <Tag />
}
```

### Alternatives Considered

currently you can make a wrapper element which just takes a children prop 

which is kinda cumbersome, and not really dynamic

like this: 

```jsx
component CustomTag(props: { tag: string, children: Component }) {
  if (props.tag === "a") {
    <a><props.children/></a>
  } else if (props.tag = "div") {
    <div><props.children/></div>
  } else if ...
}
```

### Examples & References

in react you can do 

```jsx
const Tag = "div";

return <Tag />;
```

or in svelte

```svelte
<svelte:element this="div" />
```

### Checklist

- [x] I've searched for existing feature requests
- [ ] I'm willing to help implement this feature

leonidaz:
as far as syntax maybe the following:

```jsx
let tag = track('div');

<{@tag} .... ></{@tag}> 
```

this would let the compiler know that this an html element vs a component.

whereas dynamic components are:

```jsx
let component = track(() => Component);

<@component ... >....</@component>
```

anubra266:
https://github.com/trueadm/ripple#dynamic-components

leonidaz:
This is not about dynamic components but dynamic elements

Reopening

trueadm:
I think we can probably handle this:

```jsx
let tag = track('div');

<@tag .... ></@tag> 
```

as it's done at runtime we can see it's a string and just create DOM element instead. Unfortunately, this:

```jsx
let tag = track('div');

<{@tag} .... ></{@tag}> 
```

Is a nightmare to parse and syntax highlight so we should probably avoid it like the plague.

leonidaz:
this is even better tbh, one syntax for components and tags and it seem runtime addition is super tiny.

PR #408, opened 2025-10-04
Add dynamic elements

anubra266:
Add support for dynamic elements.

Implements #266

### DIFF
--- a/packages/ripple/src/runtime/internal/client/render.js
+++ b/packages/ripple/src/runtime/internal/client/render.js
@@ -128,22 +128,22 @@ export function apply_styles(element, newStyles) {
  * @returns {void}
  */
 export function set_attributes(element, attributes) {
-	let foundEnumerableKeys = false;
+	let found_enumerable_keys = false;
 
 	for (const key in attributes) {
 		if (key === 'children') continue;
-		foundEnumerableKeys = true;
+		found_enumerable_keys = true;
 
 		let value = attributes[key];
 		if (is_tracked_object(value)) {
 			value = get(value);
 		}
-		_set_attribute_helper(element, key, value);
+		set_attribute_helper(element, key, value);
 	}
 
 	// Only if no enumerable keys but attributes object exists
 	// This handles spread_props Proxy objects from dynamic elements with {...spread}
-	if (!foundEnumerableKeys && attributes) {
+	if (!found_enumerable_keys && attributes) {
 		const allKeys = Reflect.ownKeys(attributes);
 		for (const key of allKeys) {
 			if (key === 'children') continue;
@@ -153,7 +153,7 @@ export function set_attributes(element, attributes) {
 			if (is_tracked_object(value)) {
 				value = get(value);
 			}
-			_set_attribute_helper(element, key, value);
+			set_attribute_helper(element, key, value);
 		}
 	}
 }
@@ -164,7 +164,7 @@ export function set_attributes(element, attributes) {
  * @param {string} key
  * @param {any} value
 */
-function _set_attribute_helper(element, key, value) {
+function set_attribute_helper(element, key, value) {
   if (key === 'class') {
     const is_html = element.namespaceURI === 'http://www.w3.org/1999/xhtml';
     set_class(/** @type {HTMLElement} */ (element), value, undefined, is_html);
--- a/packages/ripple/src/runtime/internal/client/render.js
+++ b/packages/ripple/src/runtime/internal/client/render.js
@@ -128,37 +128,66 @@ export function apply_styles(element, newStyles) {
  * @returns {void}
  */
 export function set_attributes(element, attributes) {
+	// Fast path: Use for...in for regular objects (99% of cases)
+	// Avoids expensive Reflect.ownKeys calls for normal elements
+	let hasAttributes = false;
+
 	for (const key in attributes) {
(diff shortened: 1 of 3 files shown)


leonidaz:
can we add tests where class and style are objects and also one where there is a spread on the element, including a ref

trueadm:
I wonder why you have all these whitespace issues? Are you using Prettier?

packages/ripple/src/runtime/internal/client/render.js
```diff
@@ -128,36 +128,65 @@ export function apply_styles(element, newStyles) {
...
 export function set_attributes(element, attributes) {
-  for (const key in attributes) {
-    if (key === 'children') continue;
-
-    let value = attributes[key];
-
-    if (is_tracked_object(value)) {
-      value = get(value);
-    }
+	const regularKeys = Object.keys(attributes);
```
trueadm: We want to avoid these expensive calls during the normal fast path where we're not dealing with dynamic elements

trueadm:
We need to be very mindful of performance overhead. The existing functions that apply attributes is very prone to exploding runtime performance if we're not careful.

anubra266:
Yeah @trueadm prettier is troublesome for me

packages/ripple/src/runtime/internal/client/render.js
```diff
@@ -128,36 +128,65 @@ export function apply_styles(element, newStyles) {
...
+
+			let value = attributes[key];
+			if (is_tracked_object(value)) {
+				value = get(value);
+			}
+			_set_attribute_helper(element, key, value);
+		}
+	} else {
+		// Use faster for...in for regular objects
+		for (const key in attributes) {
```
leonidaz: afaik, the `in` loop is slower than `of`

packages/ripple/src/runtime/internal/client/render.js
```diff
@@ -128,36 +128,65 @@ export function apply_styles(element, newStyles) {
...
-    if (is_tracked_object(value)) {
-      value = get(value);
-    }
+	const regularKeys = Object.keys(attributes);
+	const allKeys = Reflect.ownKeys(attributes);
+
+	// If Object.keys gives different results than Reflect.ownKeys, it's likely a Proxy
+	const isLikelyProxy = regularKeys.length === 0 && allKeys.length > 0;
+
+	if (isLikelyProxy) {
```
leonidaz: why is proxy special?  e.g. we can just have one `for of` loop on `attributes`

anubra266: @leonidaz attibutes is an object, to use for of, we need to do Object.keys first, we'd rather not do that when it's not neccessary i.e. it's not a proxy

trueadm: For in is by far the fastest on objects.

packages/ripple/src/runtime/internal/client/render.js
```diff
@@ -128,39 +128,66 @@ export function apply_styles(element, newStyles) {
  * @returns {void}
  */
 export function set_attributes(element, attributes) {
+	let foundEnumerableKeys = false;
```
trueadm: Use snake case

packages/ripple/src/runtime/internal/client/render.js
```diff
@@ -128,39 +128,66 @@ export function apply_styles(element, newStyles) {
...
 	for (const key in attributes) {
 		if (key === 'children') continue;
+		foundEnumerableKeys = true;
 
 		let value = attributes[key];
-
 		if (is_tracked_object(value)) {
 			value = get(value);
 		}
+		_set_attribute_helper(element, key, value);
```
trueadm: Not a fan of the underscore prefix

packages/ripple/src/runtime/internal/client/render.js
leonidaz: I meant we just call Reflect.ownKeys and do one `for of` loop, avoiding prototype props, etc. and it gets us any symboled properties and it's a lot less code.

anubra266: But @leonidaz `Reflect.ownKeys` is an overhead when not using dynamic elements

trueadm: Yeah we don’t want to make the simple case slow. This function is very hot in so many codebases that use spread. It has to be as optimal as possible 

Reflect.ownKeys is expensive to use

leonidaz: ok, i had to dig through the code, i understand what's happening now. 

so attributes is either a proxy or not, ok, yeah, this makes sense to optimize it this way.

packages/ripple/src/runtime/internal/client/render.js
```diff
@@ -128,39 +128,68 @@ export function apply_styles(element, newStyles) {
  * @returns {void}
  */
 export function set_attributes(element, attributes) {
+	// Fast path: Use for...in for regular objects (99% of cases)
+	// Avoids expensive Reflect.ownKeys calls for normal elements
+	let hasAttributes = false;
```
trueadm: Snake case please

anubra266: it's outdated

PR #415, opened 2025-10-05
fix: not unwrapping @ in effect

leonidaz:
This fixes an issue where the compiler was not applying get unwrapper for right side expressions inside effect

The new test shows where it was failing.

### DIFF
--- a/packages/ripple/tests/client/__snapshots__/compiler.test.ripple.snap
+++ b/packages/ripple/tests/client/__snapshots__/compiler.test.ripple.snap
@@ -1,0 +1,12 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`compiler success tests > compiles tracked values in effect with assignment expression 1`] = `"state.count = _$_.get(count);"`;
+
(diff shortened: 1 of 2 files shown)


packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -175,10 +177,10 @@ const visitors = {
 						add_ripple_internal_import(context);
 						return b.call('_$_.get', build_getter(node, context));
 					}
-				}
 
-				add_ripple_internal_import(context);
-				return build_getter(node, context);
+					add_ripple_internal_import(context);
+					return build_getter(node, context);
```
leonidaz, line 182: this was moved inside the if as we were unnecessarily calling these.  It now skips all the local variables, function names, parameters that never needed transformation

trueadm (approved):
Nice! Can we add a test with an update expression too?

leonidaz:
> Nice! Can we add a test with an update expression too?

@trueadm is this what you meant?

https://github.com/trueadm/ripple/pull/415/commits/8ccbda2d5cb31131eec4dcaa33e5c245bead640d

PR #419, opened 2025-10-05
fix: async with pending and regular try catch

leonidaz:


packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -241,7 +241,12 @@ export function run_block(block) {
 		active_dependency = null;
 		var res = block.fn();
 
-		if (typeof res === 'function') {
+		// Handle async functions that return a Promise
```
trueadm, line 244: We can't do this here, this is an extremely sensitive function perf wise and this completely kills performance. We need to handle this as a wrapper somewhere else

packages/ripple/src/runtime/internal/client/try.js
```diff
@@ -38,15 +38,25 @@ export function try_block(node, fn, catch_fn, pending_fn = null) {
...
 	 */
 	function move_block(block, fragment) {
-		var state = block.s;
-		var node = state.start;
-		var end = state.end;
-
-		while (node !== null) {
-			var next = node === end ? null : next_sibling(node);
+		// For blocks with DOM state (like HTML blocks), move their tracked DOM nodes
+		if (block.s !== null && block.s.start !== undefined) {
```
trueadm, line 42: These should never be `null` or `undefined`?

packages/ripple/src/runtime/internal/client/try.js
```diff
@@ -71,10 +81,10 @@ export function try_block(node, fn, catch_fn, pending_fn = null) {
 					destroy_block(b);
 				}
 				/** @type {ChildNode} */ (anchor).before(
-					/** @type {DocumentFragment} */ (offscreen_fragment),
+					/** @type {DocumentFragment} */(offscreen_fragment),
```
trueadm, line 84: Not sure why your formatter removed the white space here, but that's not right

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -1221,16 +1221,23 @@ const visitors = {
...
 		}
 
 		context.state.init.push(
 			b.stmt(
 				b.call(
 					'_$_.try',
 					id,
-					b.arrow([b.id('__anchor')], b.block(body)),
+					// Only make arrow async if there's no _$_.async wrapper (which is already async)
+					b.arrow([b.id('__anchor')], b.block(body), !has_async_wrapper && metadata.await),
```
trueadm, line 1240: Yeah this scares me, we haven't ever made any other block functions async before and doing so will cause a world of issues because DOM insertions will now be off and the runtime will have to deal with this complexity somehow, we need to be smarter here and avoid this entirely.

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -1221,16 +1221,23 @@ const visitors = {
...
 		});
 
-		if (metadata.pending) {
-			body = [b.stmt(b.call('_$_.async', b.thunk(b.block(body), true)))];
+		// If there's a pending block and the try block contains await, wrap with _$_.async
+		const has_async_wrapper = node.pending !== null && metadata.await;
+
+		if (has_async_wrapper) {
+			// Create an async arrow function that wraps the body
+			// The outer arrow (__anchor) => { ... } should NOT be async
```
trueadm, line 1229: But you made it async below?

trueadm:
Can you provide the before/after code to show what the transformations are being made please? Like the original source and the compiled client code, so it's easier to review the parts? That would be awesome

packages/ripple/src/runtime/internal/client/try.js
leonidaz: for pending blocks they are.

packages/ripple/src/runtime/internal/client/try.js
leonidaz: yeah, we should compare settings.

packages/ripple/src/runtime/internal/client/runtime.js
leonidaz: I don't think i have the full mental picture to figure this out further, without really digging into it and spending a ton of time.  Maybe it's best if you take this over?  I can be more productive in other places.

packages/ripple/src/compiler/phases/3-transform/client/index.js
leonidaz: that's where I thought we'd also use the `_$_.async` wrapper, this the case for a regular `try/catch` with an await inside or an async component.  And, basically we'd assume it's a pending with an empty string and replace it with the resolved.

packages/ripple/src/compiler/phases/3-transform/client/index.js
leonidaz: it's the outer async which is not needed here when wrapping within the `_$_.async` that has the fn parameter as an async function.

like, no async in front of `(__anchor)`

```jsx
	_$_.try(
		node,
		(__anchor) => {
			_$_.async(async () => {......
```

trueadm:
Closing this for now as we'll approach it differently with a more generic approach.

PR #420, opened 2025-10-06
Add for loop key docs

anubra266:
Add for loop key docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,7 +613,15 @@ The `for...of` loop has also a built-in support for accessing the loops numerica
 
 ```jsx
   for (const item of items; index i) {
-    <div>{item}{' at index '}{i}</div>
+    <div>{item.label}{' at index '}{i}</div>
+  }
+```
+
+You can also provide a `key` for efficient list updates and reconciliation:
+
+```jsx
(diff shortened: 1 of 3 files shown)


JakeIsMeh:
would you mind elaborating on keyed vs unkeyed usage as per the team discord? thanks

leonidaz:
We should wait until the keyed loop logic is updated.  Seems like it’s not quite doing what would be typically expected.

README.md
```diff
@@ -613,7 +613,15 @@ The `for...of` loop has also a built-in support for accessing the loops numerica
 
 ```jsx
   for (const item of items; index i) {
-    <div>{item}{' at index '}{i}</div>
+    <div>{item.label}{' at index '}{i}</div>
+  }
+```
+
+You can also provide a `key` for efficient list updates and reconciliation:
```
trueadm, line 681: We should ad a part where object reference isn't sufficient. If people are using `#{}` in their array then they likely never need to use keyed for loops, and for loops based on identity are more efficient as there's less bookkeeping.

trueadm:
Can we please get this updated and merge it?

PR #423, opened 2025-10-06
Extract LSP from VS Code extension

### LINKED ISSUE #409: IntelliJ integration

Yusyuriv:
Closes #409

This PR extracts the LSP from the VS Code extension into its own package. The idea is to use the LSP in other editors. I successfully tested it in WebStorm, Sublime Text, and Neovim. Before I change the PR from draft, I have a few questions I'd like to ask people who understand the codebase better. Probably @trueadm and @leonidaz.

1) This:
```js
const initResult = server.initialize(
	params,
	createSimpleProject([getRippleLanguagePlugin()]),
	[createRippleDiagnosticPlugin()],
);
```
seems to work in VS Code, but not in any other editor. All other editors correctly display compilation errors, but no TypeScript suggestions. I replaced `createSimpleProject` with `createTypeScriptProject`:
```js
const ts = require('typescript');

const initResult = server.initialize(
	params,
	createTypeScriptProject(
		ts,
		undefined,
		() => {
			return {
				languagePlugins: [getRippleLanguagePlugin()],
			};
		}
	),
	[
		createRippleDiagnosticPlugin(),
		...createTypeScriptServices(ts),
	],
);
```
With these changes it works everywhere, including VS Code. Is this change okay?

2) ~~The LSP was crashing with the error message `Unhandled AST node type in mapping walker: Literal` until I added a new condition to the list of skipped types in `packages/ripple/src/compiler/phases/3-transform/segments.js`:~~ This has been resolved.
```js
else if (node.type === 'Literal' || /* a long list of other types */) {
	// Primitive type keywords - leaf nodes, no children
	return;
}
```

3) The LSP seems to freak out when you make a syntax error, highlighting a lot of stuff as errors until you fix the actual error. This happens because when Ripple compilation fails, it falls back to treating the file as a regular TypeScript file, which will always fail since .ripple files aren't valid .ts files. I made it not provide any TypeScript diagnostics until Ripple compilation issue is fixed. Is this behavior good enough?
image

## Demo screenshots

WebStorm
image

Sublime Text
image

Neovim
image

## TODO
- [ ] Write installation instructions for popular editors.
- [x] Verify that VS Code extension still correctly builds. I only tested it unpacked.
> [!NOTE]
> It should correctly build once the language server is published on npm, since it uses yarn to build, and yarn doesn't work with pnpm's workspaces. It ignores them and tries to pull the package from the registry anyway.
- [ ] Someone will need to claim the package name on npm.
- [ ] Update documentation.
- [x] Potentially fix the fallback to TypeScript on compilation errors?

### DIFF
--- a/packages/ripple-language-server/LICENSE
+++ b/packages/ripple-language-server/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Dominic Gannaway
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/ripple-language-server/README.md
+++ b/packages/ripple-language-server/README.md
@@ -1,0 +1,43 @@
+# ripple-language-server
+
+Language Server Protocol (LSP) implementation for Ripple. This package provides language intelligence features for
+Ripple files and can be integrated into any editor that supports LSP.
+
+## Features
+
+- TypeScript integration via Volar
+- Ripple syntax diagnostics
+- IntelliSense and autocomplete
+- Go to definition
+- Find references
+- Hover information
+
+## Installation
+
+```bash
+npm install ripple-language-server -g
+```
+
+### Editor Integration
+
+This language server can be integrated into any editor that supports LSP:
+
+#### VS Code
+
+Use the [official extension](https://marketplace.visualstudio.com/items?itemName=ripplejs.ripple-vscode-plugin) instead.
+It uses this language server internally.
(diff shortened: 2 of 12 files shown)


leonidaz:
> 3\. The LSP was crashing with the error message `Unhandled AST node type in mapping walker: Literal` until I added a new condition to the list of skipped types in `packages/ripple/src/compiler/phases/3-transform/segments.js`:
> 
> ```js
> else if (node.type === 'Literal' || /* a long list of other types */) {
> 	// Primitive type keywords - leaf nodes, no children
> 	return;
> }
> ```

I think this good. But I wonder if this list is complete...  But I'd say let's ship it and we can always fix.

> seems to work in VS Code, but not in any other editor. All other editors correctly display compilation errors, but no TypeScript suggestions. I replaced `createSimpleProject` with `createTypeScriptProject`:
> 
> ```js
> const ts = require('typescript');
> 
> const initResult = server.initialize(
> 	params,
> 	createTypeScriptProject(
> 		ts,
> 		undefined,
> 		() => {
> 			return {
> 				languagePlugins: [getRippleLanguagePlugin()],
> 			};
> 		}
> 	),
> 	[
> 		createRippleDiagnosticPlugin(),
> 		...createTypeScriptServices(ts),
> 	],
> );
> ```
> 
> With these changes it works everywhere, including VS Code.

I think you're in the best position to make this call.  If it works, let's ship it!

leonidaz:
we should merge this, i ran into the same issues

Yusyuriv:
Still waiting on @trueadm's opinion. I'd also like to finish at least points 2 and 5 of the TODO list before merging. The rest of the TODO points can be fixed in a later PR. I kind of fixed point 5, but I'd like to hear your opinion on if this solution is good enough: when Ripple compilation fails, it stops displaying any TypeScript diagnostics at all, and instead displays the Ripple compilation error's location until it's fixed:

```diff
- // When compilation fails, use the original code as-is
- // This way positions match exactly and we can provide diagnostics on raw text
+ // When compilation fails, show where it failed and disable all
+ // TypeScript diagnostics until the compilation error is fixed
  log('Compilation failed, using original code for diagnostics');

- this.generatedCode = this.originalCode;
+ this.generatedCode = `export {};\n`;
  this.isErrorMode = true; // Flag to indicate we're in diagnostic-only mode

  // Create 1:1 mappings for the entire content
  this.mappings = [
    {
-	sourceOffsets: [0],
+	sourceOffsets: [this.errors[0]?.pos ?? 0],
	generatedOffsets: [0],
	lengths: [this.originalCode.length],
	data: {
		verification: true,
	},
    },
  ];
```

No compilation errors, TypeScript LSP works:
image

Compilation error, TypeScript LSP doesn't work until Ripple compilation error is fixed:
image

Yusyuriv:
Okay, TODO 2 and 5 done. I also came up with an alternative way to build the extension because `ripple-server-language` is not published to npm yet, and yarn ignores `workspace:*` instructions, trying to pull from npm anyway. I'll suggest it in a separate PR. Now only waiting on @trueadm

PR #427, opened 2025-10-06
fix: ssr await conditional logic

WebEferen:
### Fixed the await parsing issue in transformer. 
Changed from using a ternary conditional expression with await (which confused Rollup's parser) to using an if-statement.


trueadm (approved):
makes sense to me :) nice work

PR #430, opened 2025-10-07
chore: eslint plugin (and parser)

### LINKED ISSUE #407: [Feature]: Implement ESLint support
### Feature Category

Performance Enhancement

### Problem Statement

from feedback, people want eslint support

### Proposed Solution

plugin for eslint

### Alternatives Considered

_No response_

### Examples & References

_No response_

### Checklist

- [ ] I'm willing to help implement this feature

WebEferen:
Finally the time has come. Our own eslint plugin and parser that works like a charm with the ripple ecosystem.

image

Closes #407


JakeIsMeh:
We already have `tsdown` as a dependency for `create-ripple`, do you think you could swap `tsup` for it?

WebEferen:
> We already have `tsdown` as a dependency for `create-ripple`, do you think you could swap `tsup` for it?

@JakeIsMeh  I was more familiar with it and tsdown had some weird issues that I could not solve for some reason. If you can take a look on it, it would be awesome.

leonidaz:
Amazing work! 🙌

One question about require component export rule, I don’t quite get the use case and why it’s required?

WebEferen:
@leonidaz its just a warning highlighting - it's not an error itself. It's to inform the user that there is no default export (default exports are more performant than named ones AFAIK).

trueadm:
> @leonidaz its just a warning highlighting - it's not an error itself. It's to inform the user that there is no default export (default exports are more performant than named ones AFAIK).

Are they? What if you have a test file with not exports?

leonidaz:
One good one that I think would be helpful to add is to disallow index and key in the floor loop to be objects or arrays, at least at the top level, checking properties would be too difficult

leonidaz:
Also having to have markup in for loops, control, blocks

packages/eslint-plugin-ripple/src/rules/require-component-export.ts
```diff
@@ -0,0 +1,79 @@
+import type { Rule } from 'eslint';
+import type { ExportNamedDeclaration, ExportDefaultDeclaration } from 'estree';
+
+const rule: Rule.RuleModule = {
+  meta: {
+    type: 'suggestion',
```
WebEferen, line 6: @trueadm it's just a suggestion. Like a nitpick in that sense. I can disable / remove it but it actually have sense to keep default export suggestion. But comment with the tests itself might generate a lot of noise - will check it - that's true

WebEferen:
@leonidaz just wanted to start from something so the main parts are #/@ (unboxing and track + jsx returns). I will then update it with another parts such as for loops with elements outside of effect and for loops without elements return within the effect. But I have left it for later (it's better to have something already and extend it)

leonidaz:
Yeah, totally, should’ve prefixed that we should merge and add those would be good to add later.

Which reminds me that we should come up with our own eslint rules for our repo: snake case, indents, etc. And run as an action so you get a failure if your code formatting has issues.

leonidaz:
> @leonidaz its just a warning highlighting - it's not an error itself. It's to inform the user that there is no default export (default exports are more performant than named ones AFAIK).

I’d say let’s keep it but make it disabled by default

PR #440, opened 2025-10-08
Add 2 way bind docs

anubra266:
Add 2 way bind docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -914,6 +914,34 @@ export component App() {
 }
 ```
 
+#### Two-Way Data Binding
+
+Ripple provides utilities for two-way form binding to eliminate boilerplate:
+
+```jsx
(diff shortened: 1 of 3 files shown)


leonidaz:
I’m of the opinion that we should wait until we have more of the binders and have a chance to test it out thoroughly or create a bunch more tests.

hrutavmodha:
> I’m of the opinion that we should wait until we have more of the binders and have a chance to test it out thoroughly or create a bunch more tests.

@leonidaz 
Would you like me to help with wiring the two-way data binders in Ripple or write some of the related tests?

trueadm:
Yeah let's add the docs when we feel good about them. Plus their API has changed, and the import has changed etc since.

PR #446, opened 2025-10-09
feat(create-ripple) Allow relative path in project name prompt

qubydev:
image

- Supports relative path syntax (., ..) in project name prompts.
- Aborts if the target directory is not empty (inspired by `create-next-app`).
- Includes tests for the new behavior.

The `.` (root) is very commonly used in project name prompts, so supporting it improves usability


qubydev:
> should more tests be added?

Actually yes. specially name containing a back slash (OS dependent). I just couldn't think of how to write that test case (lol)

and a couple more.

WebEferen:
Looks good. Awesome work 👌
@qubydev - would be good to cover `..` and `../xxx` paths with test cases, to make sure it works fine.

qubydev:
> Looks good. Awesome work 👌 @qubydev - would be good to cover `..` and `../xxx` paths with test cases, to make sure it works fine.

Ok great, let me add those.

JesseKoldewijn:
Great stuff man! 🎉

trueadm:
Great stuff, let's get feedback addressed and then we can get this merged :)

qubydev:
@WebEferen @leonidaz @trueadm  I have added more test cases, can you please check?

PR #453, opened 2025-10-10
test: reorganize main package tests and fix type errors in them

Yusyuriv:
Multiple changes:

1. Long (500+ lines) tests have been split into separate files.
   
2. The setup code has been extracted from all tests. Now it runs automatically, no need to include it in each test. `describe()` `it()` straight away!
   
3. `render()`, `container`, and `error` that were previously created by the setup code locally in each test suite are now globally available in all tests.
   
   
4. vitest functions are also now globally available, no need to import them:


Yusyuriv:
Just for future reference: I explained that all tests are there. The mentioned test, for example, is in `basic.collections.test.ripple` since it was originally in `basic.test.ripple`.

PR #454, opened 2025-10-10
public api for get and set

leonidaz:
We can adjust as needed.

### DIFF
--- a/packages/ripple/src/runtime/internal/client/runtime.js
+++ b/packages/ripple/src/runtime/internal/client/runtime.js
@@ -775,7 +775,7 @@ export function get_tracked(tracked) {
  * @param {any} value
  */
 export function public_set(tracked, value) {
-	set(tracked, value, /** @type {Block} */ (active_scope));
+	set(tracked, value, safe_scope());
 }
 
 /**


packages/ripple/src/runtime/internal/client/utils.js
```diff
@@ -35,14 +35,6 @@ export function create_anchor() {
 	return t;
 }
 
-/**
- * @param {any} value
- * @returns {boolean}
- */
-export function is_positive_integer(value) {
-	return Number.isInteger(value) && /**@type {number} */ (value) >= 0;
-}
```
leonidaz, line 44: unused

packages/ripple/src/runtime/index-server.js
```diff
@@ -12,6 +12,30 @@ export function effect() {
 
 var empty_get_set = { get: undefined, set: undefined };
 
+/**
+ * @param {Derived | Tracked} tracked
+ * @returns {any}
+ */
+export function get(tracked) {
```
leonidaz, line 19: not sure if we want a public version of `get` where we'd call `safe_scope` to make sure the api is being called within the ripple context.  Otherwise, I don't see anywhere where there'd an exception thrown, so the calls will still succeed.

PR #464, opened 2025-10-12
test: fix eslint-plugin-ripple tests silently failing and add missing tests

### LINKED ISSUE #462: [Bug]: eslint plugin tests are not working
They report as successful, but actually they're silently failing.

Yusyuriv:
Closes #462.

As I mentioned in the issue, eslint plugin tests were silently failing. I fixed that. Now the tests are working as expected, revealing several other issues fixing which is beyond the scope of this PR.

The problem with tests failing consisted of several layers:

1. I assume `ruleTester.run()` already uses `describe()` and `it()` under the hood, and `it()` can't be nested. I removed our `describe()` and `it()`, executing `ruleTester.run()` directly.
2. None of the tests specified `eslint-parser-ripple` as the parser when testing rules, so it was failing to parse the Ripple code in tests.
3. Specifically `require-component-export` test was failing because its rule was not visiting `Component` AST node.

I also added missing tests:
1. `require-component-export`
2. `unbox-tracked-values`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,10 @@ jobs:
         working-directory: ./packages/create-ripple
         run: pnpm build
 
+      - name: Build eslint-parser-ripple
+        working-directory: ./packages/eslint-parser-ripple
+        run: pnpm build
+
       - name: Run tests
         run: pnpm test
 


trueadm:
Looks like the CI is still failing?

Yusyuriv:
The CI is now failing because eslint plugin tests now actually run instead of failing silently like before. @WebEferen is looking into this.

PR #466, opened 2025-10-12
test: add tests for parts of `ripple/src/utils`

Yusyuriv:
Added 112 tests. I didn't add tests for `ast.js` and `builders.js` because I don't know what and how to test there.

### DIFF
--- a/vitest.config.js
+++ b/vitest.config.js
@@ -60,6 +60,14 @@ export default defineConfig({
 				},
 				plugins: [ripple()],
 			},
+			{
+				name: 'ripple-utils',
+				test: {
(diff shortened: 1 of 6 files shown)


vitest.config.js
```diff
@@ -60,6 +60,14 @@ export default defineConfig({
 				},
 				plugins: [ripple()],
 			},
+			{
+				name: 'ripple-utils',
```
trueadm, line 64: What's this for?

Yusyuriv: `escaping.js` is used by both client and server, so I was not sure where I should put the tests: `ripple-client` or a separate test project (which I did with `ripple-utils`). Should I move them to `ripple-client`?

PR #474, opened 2025-10-13
fix: ambient declare types for the function overload

WebEferen:
Added ambient declare types that allows developers to use function overload.

### DIFF
--- a/packages/ripple/tests/client/function-overload-import.ripple
+++ b/packages/ripple/tests/client/function-overload-import.ripple
@@ -1,0 +1,10 @@
+export function test(arg: string): string
+export function test(arg: number): string
+
+export function test(arg: string | number): string {
+  if (typeof arg === 'string') {
+    return arg;
+  }
+
+  return arg.toString();
+}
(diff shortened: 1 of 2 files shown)


packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -1971,6 +1993,15 @@ export function transform_client(filename, source, analysis, to_ts) {
 		sourceMapSource: path.basename(filename),
 	});
 
+	// Post-process TypeScript output to remove 'declare' from function overload signatures
+	// Function overload signatures in regular .ts files should not have 'declare' keyword
+	if (to_ts) {
```
leonidaz, line 1998: I wonder how this will affect the volar mappings for the language server

WebEferen: It is the post-effect which only applies to the JS itself. AFAIK Volar mappings are strictly bound to the Typescript, so the flag to_ts should not be triggered). But anyway worth checking

WebEferen: This is vital as without of this post-effect, we will be getting this typescript error

PR #486, opened 2025-10-14
fix: vscode extension ripple fallback missing

WebEferen:
Fixed the priority for the ripple lookup within the vscode extension. Now it firstly search for the users managed ripple workspace, then it uses pre-bundled ripple version bundled with vscode extension.


leonidaz:
@WebEferen thanks, but i've been working on this already.

WebEferen:
@leonidaz closing the PR then.

PR #496, opened 2025-10-15
build: Simplify build commands for VS Code extension; let tests use eslint plugin without building it

Yusyuriv:
* Removed esbuild in VS Code extension, replaced it with tsdown.
* `main` property in `eslint-plugin-ripple/package.json` now points to `src` so tests can be run without building it first.
* Simplified file copying scripts in VS Code extension since all they did is possible to do with just rsync.

### DIFF
--- a/packages/ripple-vscode-plugin/package.json
+++ b/packages/ripple-vscode-plugin/package.json
@@ -21,8 +21,8 @@
     "directory": "ripple-vscode-plugin"
   },
   "scripts": {
-    "copy-ripple": "rsync -aL --del --mkpath --exclude=node_modules/ripple --exclude=.ignored* node_modules/ripple/ dist/node_modules/ripple/",
-    "copy-ts": "rsync -aL --del --mkpath node_modules/typescript/ dist/node_modules/typescript/",
+    "copy-ripple": "mkdir -p dist/node_modules && rsync -aL --del --exclude=node_modules/ripple --exclude=.ignored* node_modules/ripple/ dist/node_modules/ripple/",
+    "copy-ts": "mkdir -p dist/node_modules && rsync -aL --del node_modules/typescript/ dist/node_modules/typescript/",
     "build": "tsdown",
     "dist": "perl -pi -e 's/\"main\": \"src\\/extension.js\"/\"main\": \"dist\\/extension.js\"/' package.json",
     "src": "perl -pi -e 's/\"main\": \"dist\\/extension.js\"/\"main\": \"src\\/extension.js\"/' package.json",


packages/eslint-plugin-ripple/package.json
```diff
@@ -28,10 +28,12 @@
   },
   "homepage": "https://ripplejs.com",
   "scripts": {
-    "build": "tsdown",
-    "prepack": "tsdown",
-    "test": "vitest run",
-    "test:watch": "vitest"
+		"dist": "perl -pi -e 's/\"main\": \"src\\/index.ts\"/\"main\": \"dist\\/index.js\"/' package.json",
+		"src": "perl -pi -e 's/\"main\": \"dist\\/index.js\"/\"main\": \"src\\/index.ts\"/' package.json",
```
leonidaz: indents

packages/eslint-plugin-ripple/package.json
```diff
@@ -28,10 +28,12 @@
...
+    "build": "pnpm dist && tsdown",
+    "prepack": "pnpm dist && tsdown",
```
leonidaz: we need to run `&& pnpm src` at the end to restore it back for both

packages/eslint-plugin-ripple/package.json
```diff
@@ -28,10 +28,12 @@
...
+    "test": "pnpm src && vitest run",
+    "test:watch": "pnpm src && vitest"
```
leonidaz: these can just stay as is, since we should be pointed at src always, except for a super brief moment during build and pack

packages/ripple-vscode-plugin/package.json
```diff
@@ -21,17 +21,16 @@
...
-    "build-extension": "esbuild src/extension.js --bundle --platform=node --target=node20 --outfile=dist/extension.js --external:vscode --external:typescript --external:ripple && pnpm copy-ts && pnpm copy-ripple",
-    "build-server": "esbuild src/server.js --bundle --platform=node --target=node20 --outfile=dist/server.js --external:typescript",
+    "copy-ripple": "rsync -aL --del --mkpath --exclude=node_modules/ripple --exclude=.ignored* node_modules/ripple/ dist/node_modules/ripple/",
+    "copy-ts": "rsync -aL --del --mkpath node_modules/typescript/ dist/node_modules/typescript/",
+    "build": "tsdown",
     "dist": "perl -pi -e 's/\"main\": \"src\\/extension.js\"/\"main\": \"dist\\/extension.js\"/' package.json",
     "src": "perl -pi -e 's/\"main\": \"dist\\/extension.js\"/\"main\": \"src\\/extension.js\"/' package.json",
-    "package": "pnpm dist && pnpm copy-ts && pnpm copy-ripple && vsce package --no-dependencies --out=ripple-vscode-plugin.vsix; pnpm src",
+    "package": "pnpm dist && pnpm copy-ts && pnpm copy-ripple && vsce package --no-dependencies --out=ripple-vscode-plugin.vsix && pnpm src",
```
leonidaz: we should keep the semicolon so that if anything before it fails, the second still runs.

Update: I just ran into this actually when i attempted to build the package as rsync with `--makepath` failed

packages/ripple-vscode-plugin/package.json
```diff
@@ -21,17 +21,16 @@
     "directory": "ripple-vscode-plugin"
   },
   "scripts": {
-    "copy-ripple": "mkdir -p dist/node_modules && rm -rf dist/node_modules/ripple && rsync -aL --exclude='node_modules/ripple' node_modules/ripple/ dist/node_modules/ripple/",
-    "copy-ts": "mkdir -p dist/node_modules && rm -rf dist/node_modules/typescript && cp -RL node_modules/typescript dist/node_modules/typescript",
-    "build-extension": "esbuild src/extension.js --bundle --platform=node --target=node20 --outfile=dist/extension.js --external:vscode --external:typescript --external:ripple && pnpm copy-ts && pnpm copy-ripple",
-    "build-server": "esbuild src/server.js --bundle --platform=node --target=node20 --outfile=dist/server.js --external:typescript",
+    "copy-ripple": "rsync -aL --del --mkpath --exclude=node_modules/ripple --exclude=.ignored* node_modules/ripple/ dist/node_modules/ripple/",
+    "copy-ts": "rsync -aL --del --mkpath node_modules/typescript/ dist/node_modules/typescript/",
```
leonidaz: remove `--mkpath` on both - it's not an option that rsync has but the rest is good 👍

packages/ripple-vscode-plugin/package.json
```diff
@@ -21,17 +21,16 @@
...
+    "build": "tsdown",
     "dist": "perl -pi -e 's/\"main\": \"src\\/extension.js\"/\"main\": \"dist\\/extension.js\"/' package.json",
     "src": "perl -pi -e 's/\"main\": \"dist\\/extension.js\"/\"main\": \"src\\/extension.js\"/' package.json",
-    "package": "pnpm dist && pnpm copy-ts && pnpm copy-ripple && vsce package --no-dependencies --out=ripple-vscode-plugin.vsix; pnpm src",
+    "package": "pnpm dist && pnpm copy-ts && pnpm copy-ripple && vsce package --no-dependencies --out=ripple-vscode-plugin.vsix && pnpm src",
     "install-package": "code --install-extension ripple-vscode-plugin.vsix",
-    "build-and-package": "pnpm build-extension && pnpm build-server && pnpm package",
-    "build-package-and-install": "pnpm build-extension && pnpm build-server && pnpm package && pnpm install-package",
-    "build-package-install-and-run": "pnpm build-extension && pnpm build-server && pnpm package && pnpm install-package && code"
+    "build-and-package": "pnpm build && pnpm package",
```
leonidaz, line 31: i guess they have to use `eval` - I couldn't find an option for this:

```
[EVAL] Warning: Use of `eval` function is strongly discouraged as it poses security risks and may cause issues with minification.
    ╭─[ ../../node_modules/.pnpm/@volar+language-server@2.4.23/node_modules/@volar/language-server/node.js:47:22 ]
    │
 47 │     const _require = eval('require');
    │                      ──┬─  
    │                        ╰─── Use of `eval` function here.
────╯

[EVAL] Warning: Use of `eval` function is strongly discouraged as it poses security risks and may cause issues with minification.
    ╭─[ ../../node_modules/.pnpm/@volar+language-server@2.4.23/node_modules/@volar/language-server/lib/features/editorFeatures.js:65:30 ]
    │
 65 │             const _require = eval('require');
    │                              ──┬─  
    │                                ╰─── Use of `eval` function here.
────╯
```

leonidaz, line 31: they're including `path-browserify` - couldn't find an option to turn it off


Image

packages/ripple-vscode-plugin/package.json
Yusyuriv: I took it directly from `rsync --help`:
```
--mkpath                 create destination's missing path components
```
Without this option it will fail if the destination doesn't exist.

packages/ripple-vscode-plugin/package.json
Yusyuriv: It's a direct dependency of `@volar/language-server` and doesn't seem to work without it. Not sure why `tsdown` doesn't bundle this one, but the extension doesn't work without this, and I don't see any harm in having this file in there.

packages/ripple-vscode-plugin/package.json
Yusyuriv: I encountered an issue with `;` on Windows, but I think I found a solution.

packages/eslint-plugin-ripple/package.json
```diff
@@ -28,8 +28,10 @@
   },
   "homepage": "https://ripplejs.com",
   "scripts": {
-    "build": "tsdown",
-    "prepack": "tsdown",
+    "dist": "perl -pi -e 's/\"main\": \"src\\/index.ts\"/\"main\": \"dist\\/index.js\"/' package.json",
+    "src": "perl -pi -e 's/\"main\": \"dist\\/index.js\"/\"main\": \"src\\/index.ts\"/' package.json",
+    "build": "pnpm dist && tsdown && pnpm src || pnpm src",
```
Yusyuriv, line 33: This makes sure `pnpm src` runs both on success and on fail, and it works on Windows.

packages/ripple-vscode-plugin/package.json
leonidaz: yeah, but it's for linux, mac has the BSD version.

let's keep mkdir as that solution is cross-platform, assuming WSL

packages/ripple-vscode-plugin/package.json
leonidaz: ah, i thought it was the tsdown that added that since i hadn't seen it before.

packages/eslint-plugin-ripple/package.json
leonidaz: this works but interesting, so `;` doesn't work on windows?

Yusyuriv: It does usually:
```
> echo 1; echo 2
< 1
< 2
```
But when I use pnpm to run a similar command, I assume npm/pnpm runs it differently, it considers `;` and everything after it to be a part of the previous command.
```json
"echo": "echo 1; echo 2"
```
```
> pnpm echo
< 1; echo 2
```
Funnily enough, `&&` and `||` don't work in PowerShell, but the same npm/pnpm magic makes it actually run correctly.

leonidaz (approved):
👍

PR #508, opened 2025-10-23
fix missing prop descriptors

### LINKED ISSUE #448: [Bug]: Getting compilation error when using @tracked_var inside ts file
### Issue Category

Compilation Bug

### Bug Description

I come across this scenario when I created a `.ts` store file and using effect it gives compilation error

```jsx 
import { effect, track } from "ripple"

// countStore.ts
export function useCount() {
  const count = track(1)
  effect(() => {
    console.log(@count)
  })
  return { count }
}

```

```jsx
import { useCount } from './countStore'; //using countStore.ts
import { effect } from "ripple"

export component App() {
  const { count } = useCount()
  effect(() => {
   const id = setInterval(() => {
      @count++;
    }, 1000);

    return () => clearInterval(id)
  })
  <h1>{@count}</h1>
}
```

getting compilation error
```bash
[plugin:vite:esbuild] Transform failed with 1 error:
C:/Workspace/my-ripple-app/src/countStore.ts:7:22: ERROR: Expected "class" but found ")"
```

### Reproduction Steps

1. create a sample app using latest ripple version
2. create countStore.ts and copy the copy
3. update App.ripple and copy App component code

### Environment

```markdown
- latest ripple version
```

### Additional Context

But when I used `.ripple` file as store it works, like below
```jsx
import { effect, track } from "ripple"

// countStore.ripple
export function useCount() {
  const count = track(1)
  effect(() => {
    console.log(@count)
  })
  return { count }
}

```
```jsx
import { useCount } from './countStore.ripple';
import { effect } from "ripple"

export component App() {
  const { count } = useCount()
  effect(() => {
   const id = setInterval(() => {
      @count++;
    }, 1000);

    return () => clearInterval(id)
  })
  <h1>{@count}</h1>
}
```

### Checklist

leonidaz:
fixes #448 

- We need to get the actual descriptors otherwise we lose the getters.

### DIFF
--- a/packages/ripple/tests/client/composite/composite.props.test.ripple
+++ b/packages/ripple/tests/client/composite/composite.props.test.ripple
@@ -1,4 +1,4 @@
-import { track, effect, flushSync } from 'ripple';
+import { track, trackSplit, effect, flushSync, type Props } from 'ripple';
 
 describe('composite > props', () => {
 	it('correctly handles default prop values', () => {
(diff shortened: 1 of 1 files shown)


packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -922,11 +922,7 @@ export function spread_props(fn, block) {
 				const obj = get_derived(computed);
 
 				if (key in obj) {
-					return {
-						enumerable: true,
-						configurable: true,
-						value: obj[key],
-					};
+					return Object.getOwnPropertyDescriptor(obj, key);
```
trueadm: We have an import we can use instead of doing this “own_property_descriptor”

trueadm (approved):

PR #509, opened 2025-10-23
Bump Ripple package versions

Aniketkhote:
- Update Node.js requirement from 18.0.0 to 20.0.0 in README documentation
- Update Ripple packages in basic template to latest versions (0.2.140)
- Improve code formatting and consistency across project-creator and validation modules
- Clean up import order in project-creator.js


leonidaz:
I think for the basic template we should just change all ripple packages to use `latest` for package versions.

just gotta make sure that `create-ripple` can handle that.

WebEferen:
We'd just need to bump the basic template version(s) to latest ones. Other changes are not that much important now as they will be a subject to change in near future. @leonidaz

PR #515, opened 2025-10-28
feat: eslint rule for the stringed version of keys in for..of statements

WebEferen:
Added an eslint rule for making sure that destructuring objects can be used in stringified form as the keys.
Closes https://github.com/trueadm/ripple/issues/433


leonidaz:
this issue was only for the `key`, not the loop itself.

WebEferen:
The new pr will come with corrected version of the eslint rule alongside with the tests and small fixes to previous ones.

PR #516, opened 2025-10-28
feat: added missing multiple styled component test

WebEferen:
Added a test case that ensures that stylesheets are correctly applied across the files.
Closes https://github.com/trueadm/ripple/issues/489


leonidaz (commented):
we have these tests in the other pr with :global, so doesn't seem like we need this pr.

WebEferen:
@leonidaz Actually this PR should be the main one for the missing tests for the nested styling, the latter one was about the global selector styles (but my stash went wrong).

PR #522, opened 2025-10-30
added default exports in vite-plugin-ripple

chandani-rathi:
after taking the latest code changes, my vite plugin code is breaking due to missing default exports in vite-plugin-ripple


chandani-rathi:
@leonidaz : please review.

leonidaz:
thank you!

PR #527, opened 2025-11-03
alternative props spread to keep nested component reactivity

### LINKED ISSUE #525: [Bug]: Props spread and trackSplit is not working correctly together
### Issue Category

Behavioral Bug

### Bug Description

I am not sure if this is an issue, but props spread and trackSplit are working correctly together.

I know I can pass the tracked value, but in some simple cases, it works correctly without passing tracked value only when doing some complex or nest component logic it breaks.

See below example.
```jsx 
import { trackSplit, track } from "ripple";

  const count = track(0);
  <CounterWrapper {@count} up={() => @count++} down={() => @count--} /> // <-- this is not working; counter value is not updating in ui
  <hr />
  <Counter {@count} up={() => @count++} down={() => @count--} /> // <-- this is working
}

component CounterWrapper(props){
  <div>
    <Counter {...props} />
  </div>
  <style>
    div {
      display: flex;
      flex-direction: column;
      text-align: center;
      gap: 10px;
      margin: 20px;
    }
  </style>
}

component Counter(props){
  const [count, up, down, rest] = trackSplit(props, ["count", "up", "down"]);
  <button onClick={() => @up()}>{"UP"}</button>
  <span {...@rest}>{` Counter:   ${@count}   `} </span>
  <button onClick={() => @down()}>{"DOWN"}</button>
}

```

### Reproduction Steps

[Playground](https://www.ripplejs.com/playground?v=0.2.151&title=Hello+World#config=code%2FN4IgLglmA2CmIC4QAlbWgewAQHUMCdoATEAGhCNgGcBjfCAB0gwDtEQyQALWAQxKQAeALawwvLDS698VMQF4AOiACqAFQBiAWgAcyrAHoAfIpYixElr1FKQANwiwA7gwJh9NVmFgswtpxBEYFzylA40sFoBQVykWBAsUBC80Fq0KbDyAIwAdAAM%2BsacXGDC0ACCYGCy7NC8LADmtj4edVRUtsqc4g1UiADaALrkvDSQdrAAokRQBOy09EycwjIA1gCuDIigdY3rvA3wSCVlnJ6%2BPmDsgguMYFhgAJ4MmcrCGETrcMomLIqQwlc%2BHuwCw73WviwAF8sAAzfAYYRYADkiwYcGRAG5TP8IIC3FgAFRYXhULC3JhwhFI5Q5AwUsA5NHfEDYv4sLCSVhUe6wAAeQLAZPkWAA8gAjABWsDGOVWsEeVAAFAyAJRsznnHlYcoMBhYEX8wVUHJwRrBA3yEVZHHuMAAfnJdDu-SNbio-Tyg0Gtv%2BCCwKudTBylFhvC%2B9wAPpGnYtGbr9dHYy63cCTbCEkQlbADUYHs9YBhYcmmK7BpaRcjYRCxhBWMisKrBurTJyIMWlU8XkWdXqsABCK0o6ssWv11VYYCtzkPLgIpxYFjOLCTfAI-BK5EAOWwdhSgSwqfusIwEKIyJbHOh0-BviVCbiU6vnPE%2BEOYH9RAwNHWol8OXFD5HhyXg9R8IgAGEuAgYglS-H8-0ZOg%2BG8SY4EQzcZjsC9VVIacoUvQR6SDMAjBAKFyB5R44G2EBdgafZDnYGh2jOLxLnYYpAkoFgIPY3x2EAohHknOEvC0MNhBgx5-S0UD0UiKhFW8YQ4gAIWgBJVgAWVGABlZTYGEDQvDiZE9NgBoMBzFQAElkTiAAlDBALADA4lFPlHkOFg4hUcUITAdY4gg%2BpXzQaAzNFF4OT0%2BoqAclFUGgCZIBoCQt1gdZYESqh4rSWB6FhTEsHFUZVgaBEz39LgqGgJU8jiPIAFI4iyHRmvVLlMHwf0AGJYUG6FyMokjaPoxijhAZl4HIc5vAEpA8UFUTqnKvT0SgOI1poVZoSpREsGUGblDZUwjywUNw2ge5PHxJdIQTJUJyfTVuVu09IRFHbVgay9OUEPjAsKnB8HkwrJwAAU8QKYU2eRgGe3MsGhz6wAAanRmEvycFgEaR%2BQ81RwKtC0GFjEMAwsEEUnZwgMl6cXDB7icAhVgSBoSphi58CwPdoGy%2BIyRYZmsE2IheEgRp4g5dYIGnQQ50MX4AaBnmoe5sA4YYfGJ0JlHNcx7GMFx3XkeJ3xSfJvMDCpmmtDphmyVZ-B2caUwoRxUdEVcB77jV7x8FB8GNwYBEGCoVVXuprCVZnQG0Yh4AchTsOMAj62FYMWOFaouA485LDJ2nGdLvp9FeBkuE4D5DVS%2Br-ktBmfAZWYFh-U8AXhBYOvS%2B8PkwDkzSGnbyRLkK3uZwaUD-SyPIGFrkuZxWN8En9AAmefF%2Bfa8ryIvPYF%2BT32Tu33LiwAPCqVNOI6j6ctXufpNbiTY4hx3ysBbnly2%2BsHdo2zSYBr7hyoHEfoyhNbKDiMoTYUCjoUBNn8EAzY66CAClUVgWBWAQU0rtM2%2BtIabGelCIwU5VAAAVlBQiIugtyLA443AYPUScKcciQy-lrUhAADC%2BideozgACTAAtlrGcXCYT7yYfQhWtDMHYNwasfBRN37ENIcoAAIqKHAW4qE0PWBg6RLBj4jRAAfKgPAxB9AQEMUacYrE2JACxKg5Cv5iE4uQNOER2gEHscMRx6weSIgslUDmVjQAzVohMWQdY2BIDyDkdeOQsgAFYsjkQoiAZa7ptgZK7NQHJ5BvA8jCXReoDEDhTTyQyNiFxFocAyVEqgMT2AABYADs3QMAYGgCUnwvBxRwAECAFI0BOCjHGFNLU3TZqmPEEFKxIB04%2BHSUAA)

### Environment

```markdown
- latest ripple 0.2.151
```

### Additional Context

_No response_

### Checklist

leonidaz:
closes #525 

Alternative for spreading props handling that allows keeping reactivity via intermediate components and usage of trackSplit()

- instead of performing an actual destructuring of props that are spread, it keeps the props intact.
- if a component instance has explicitly specified props and spreads, an array of objects is produced by the compiler.  Otherwise, it's a single object.
- The order of prop evaluation was left intact, e.g. when a prop in spread can overwrite the explicit prop or a prop in another spread or vice versa where the explicit prop provided after spread overwrite the prop in the spread.
- The `proxy_tracked` was renamed to `proxy_props`, as we no longer need to use `derived` on the spread / props for the `spread_props` function.
- The `spread_props` function was left purely for semantics.  We could use the `proxy_props` directly but the former better reflects the intent.

Pros and Cons:

Pros:
- Lazy evaluation. Before, the props were destructured `{...props}` which would cause evaluation of all props right away for any one property access.  Now, only the requested property gets evaluated.
- The props reactivity for `trackSplit()` usage stays intact when spread props are passed through via any number of component layers
- No need to use derived for the proxy since no destructuring is actually happening and no need to check the derived's dirty flag.  The reactivity encapsulation works via just a thunk.

Cons:
- There is slightly more time spent finding the property requested as the proxy needs to loop through the array of objects vs just dealing with one object.  
  - However, typically the number of objects should be very few, especially if a spread is placed last in a component call, there would be only 2. 
  - This extra time should also be offset or even dwarfed by the previous cost of evaluating the whole destructured object for every property access.
  
Further optimizations:
- If prop spreads are always evaluated last and always overwrite any explicitly provided props, then the number of objects in the resulting array should be just 2 with one spread: 1 for explicit and 1 for the spread.  Each additional spread would still add an object to the array.

### DIFF
--- a/packages/ripple-compat-react/src/index.js
+++ b/packages/ripple-compat-react/src/index.js
@@ -5,7 +5,15 @@ import { jsx, jsxs, Fragment } from 'react/jsx-runtime';
 import { useSyncExternalStore, useLayoutEffect, useRef, useState } from 'react';
 import { createPortal } from 'react-dom';
 import { createRoot } from 'react-dom/client';
-import { branch, with_block, proxy_tracked, set, render, tracked } from 'ripple/internal/client';
+import {
+	branch,
+	with_block,
+	proxy_props,
+	set,
+	render,
+	tracked,
+	get_tracked,
+} from 'ripple/internal/client';
 import { Context } from 'ripple';
 
 /** @type {Tsx} */
@@ -138,7 +146,7 @@ export function Ripple({ component, props }) {
 		const anchor = document.createTextNode('');
 		const block = get_block_from_dom(span);
 		const tracked_props = (tracked_props_ref.current = tracked(props || {}, block));
-		const proxied_props = proxy_tracked(/** @type {any} */ (tracked_props));
+		const proxied_props = proxy_props(() => get_tracked(tracked_props));
 		frag.append(anchor);
 
 		const b = with_block(block, () => {
--- a/packages/ripple/src/compiler/phases/3-transform/client/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/client/index.js
@@ -1240,29 +1240,50 @@ const visitors = {
 			// We're calling a component from within svg/mathml context
 			const is_with_ns = state.namespace !== DEFAULT_NAMESPACE;
 
+			let object_props;
+			if (is_spreading) {
+				// Optimization: if only one spread with no other props, pass it directly
+				if (props.length === 1 && props[0].type === 'SpreadElement') {
+					object_props = b.call('_$_.spread_props', b.thunk(props[0].argument));
+				} else {
+					// Multiple items: build array of objects/spreads for proper merge order
+					const items = [];
+					let current_obj_props = [];
+
+					for (const prop of props) {
+						if (prop.type === 'SpreadElement') {
+							// Flush accumulated regular props as an object
+							if (current_obj_props.length > 0) {
+								items.push(b.object(current_obj_props));
+								current_obj_props = [];
+							}
+							// Add the spread argument directly
+							items.push(prop.argument);
+						} else {
+							// Accumulate regular properties
+							current_obj_props.push(prop);
+						}
+					}
+
+					// Flush any remaining regular props
+					if (current_obj_props.length > 0) {
+						items.push(b.object(current_obj_props));
+					}
+
+					object_props = b.call('_$_.spread_props', b.thunk(b.array(items)));
+				}
+			} else {
+				object_props = b.object(props);
+			}
 			if (metadata.tracking) {
-				const shared = b.call(
-					'_$_.composite',
-					b.thunk(visit(node.id, state)),
-					id,
-					is_spreading
-						? b.call('_$_.spread_props', b.thunk(b.object(props)), b.id('__block'))
-						: b.object(props),
-				);
+				const shared = b.call('_$_.composite', b.thunk(visit(node.id, state)), id, object_props);
 				state.init.push(
 					is_with_ns
 						? b.call('_$_.with_ns', b.literal(state.namespace), b.thunk(shared))
 						: b.stmt(shared),
 				);
 			} else {
-				const shared = b.call(
-					visit(node.id, state),
-					id,
-					is_spreading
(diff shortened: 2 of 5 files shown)


trueadm (approved):
Awesome stuff

PR #531, opened 2025-11-05
Fix compilation error when rendering empty string as @children

### LINKED ISSUE #530: [Bug]: Compilation error when rendering empty string as @children inside component
### Issue Category

Compilation Bug

### Bug Description

see below example:
```jsx
  <Button>{''}</Button>
}

component Button({ children }){
  <@children />
}
```
got exception
```bash
Error: Uncaught TypeError: Cannot read properties of null (reading 'cloneNode')

```

### Reproduction Steps

[Playground](https://www.ripple-ts.com/playground?v=0.2.152&title=Counter#config=code%2FN4IgLglmA2CmIC4QGED2BXAdmWAnEANCACawDOAxrhAA6SqaIiEgAWsAhsUwDwC2sMBwAEFVh1xlBAXgA6IAKoAVAGIBaABzzhAegB8szP0EjMHAXJAA3CLADuNVLjDaKDHNkt2IxMK2mkNhSwat6%2BrATCEJhQEBzQapTxsNIAjAB0AAza%2BiysYHzQAIJgYJJM0ByYAOaWsJiulWRklvIsQtVkiADaALpEHBSQVrAAosRQTkyU1HQsfBIA1ug0iKCVNegc1fBI%2BYUsbtj1YLwztGDCYACeNCnyfKjE6HDyBg2QfI7OwsDCj1hLgBfYQAM1wqD4wgA5LMaHBoQBuQyyT7fS4AKmEHDIwnOdDBEKh8nSOnxYHScNeIGRDUwwlEDDIl1gAA90bjpMIAPIAIwAVrAhulFrBrmQABTkgCUtIZR2ZwiKNBowi5bI56TgNT8aukXNSKJcYAA-HiqBduhqnGAyN1Mr1ekbUQhhFKLXR0qRQRwXpcAD7%2B82zCnK1WB4OW63OMjpUHRYgS2BqvRXW6wVCgyN0K29PVc6GgrBDCAMaHCaW9WWGBkQLMSm53TNKlXCACE%2BphRcwJbL0t%2BNYZV1YELswkw9mEo1wENwEuhADlUMIrPEfMJo5dQRhMMRodX6cIgYOAdgJWHIsBBwyhLgdmBXcRUBR0AJsOleU9rukOCr6sRkFYCBoETJ8XzfCkqE4HBRjgCD5wmKx92lAhByBA8eDJD0wD0EAgSIZlrjgNYQA2aoth2JgKGaQ53BOJg8h8UhMDQY5sCYT9iGuX4wXcNQfT4YDrldNRf3hEIyHFHA%2BEiAAhaBokWABZQYAGUpNgPgVHcSJoVU2BqlQZMFAASWhSIACVUE-MBUEiblWWuHZMEiBReUBdBImQKpb1gaBoF07k7npVSqjIcyYQACT8kZIAoEQF1gdBYAisgwsSPA60RBkgTwgjsJIsiKN2EAqXgIgjg8U4kE3YRvV9aBLjcL4GBOFsaAlfsr0PHhZPQUoGD0YBoWhIFML6gbMHeY86WaxwJ2wYQJtszAJT%2BMRgOIXB6iPaVuoZHgAAENpA7b6X0QxcvwkBCLgMh2EELoED6fKQyel6QGosgAAVtqkarmCIGgIWCZonHe-pPvQZlIX00pok6EiypIkZJFLRgkEydIACZ0lSABWbG8OuiAWpjNZrsbcgKaIHBmSe9YqnI7YSqp8laLYgG8usPAyHRpgABYAHZ2lQVBoAZkB6g4Xk4G4JB4mgFhBmGEqFXF8qbqEMBoaYVBguJoA)

### Environment

```markdown
- latest ripple 0.2.152
```

### Additional Context

_No response_

### Checklist

anubra266:
Fixes #530

### DIFF
--- a/packages/ripple/tests/client/basic/basic.components.test.ripple
+++ b/packages/ripple/tests/client/basic/basic.components.test.ripple
@@ -275,7 +275,9 @@ describe('basic client > components & composition', () => {
 		}
 
 		component App() {
+			let text = '';
 			<Button>{''}</Button>
+			<Button>{text}</Button>
 		}
 
 		expect(() => {


packages/ripple/src/runtime/internal/client/template.js
```diff
@@ -67,11 +67,14 @@ export function template(content, flags) {
 	var is_comment = content === '<!>';
 	var has_start = !is_comment && !content.startsWith('<!>');
```
leonidaz, line 68: I think we should just let the flow play out and mark that this content doesn't have a start.
```suggestion
	var has_start = !is_comment && content.length && !content.startsWith('<!>');
```

trueadm:
I'm not sure we should even be solving this in the template? that seems like it's far too late to handle, no? If the template is only a single string, then we should probably have a fast-path for `document.createTextNode` like Svelte 5 does.

leonidaz:
I was thinking that because a variable could be used as an empty string that we need to handle it at runtime.

But, i just checked and if there is a variable we use one space ``` ``` in the template. 

So, yeah, we can detect in the compiler that it's a literal and instead of creating a `template` call, we print `create_text` (svelte uses `text` that calls create_text) and pass the string to it.

leonidaz:
actually, even if it's a variable, we should use `create_text`.  as long as there is no markup we can avoid using `template`.

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -2274,7 +2274,13 @@ function transform_children(children, context) {
 					}
 				} else if (normalized.length === 1) {
 					if (expression.type === 'Literal') {
-						state.template.push(escape_html(expression.value));
+						if (expression.value === '') {
```
trueadm, line 2277: Does the value have to be empty? Can't it be any value in theory?

leonidaz: yeah, it shouldn't matter, for any value we should do this.

anubra266: I don't understand @trueadm

trueadm: we should do this in all cases of `normalized.length === 1` right? We just need to ensure we `escape_html` the expression if there is one

PR #561, opened 2025-11-13
Event object

leonidaz:
- Allows event handlers to be objects with options like the addEventListener with 2 additional ones
  - handleEvent() - handler
  - custom: true - keeps the letters casing as specified on the event
  - delegated: false - opts out of delegation, even for events that are typically delegated by ripple
  - once: true - fires once as per the the js spec
  - signal - abort signal as per the the js spec
  - passive - as per the js spec
  - capture - capturing phase as per the js spec
- The events with the `Capture` suffix are deprecated, use object with `capture: true`
- The events have to start with `on` but the next letter doesn't have to be capitalized to accommodate custom events.  For browser-defined events, we still recommend to stick to the camel case but they will work with whatever case and will be converted to lower case like before, unless you use `custom: true`.
- Makes the event handler (object or function) shallowly reactive, it wasn't actually working before, thus inconsistent Ripple behavior
  - if there is no tracking at parse / compile time, then the static version of the `event` creation is used.
- Fixes a few bugs along the way where
  - `on` was supposed to work with delegated
  - cleanup of removed attributes was missing
  - a couple other minor issues
- The delegation for events that use a function as a handler inside the component are statically analyzed like before.
- Other delegated events outside of component or objects can still be delegated at runtume if they're delegatable, by using the exact same mechanism as for the statically analyzed.
- Currently, if event handler is an object with any of the following options, the event will not be delegated: `capture: true`, `once: true`, `passive: true` (unless it's the common 4 passive ones, e.g. 'touchstart'), `signal`, `delegated: false`.
  - However, we can easily make `once`, `signal` also delegated
- Custom events are delegated by default, to opt out, specify `delegated: false`
- Types updated to take either a function or an object for event attributes.
- Created `render_event` as a runtime function but we can keep it in the generated code - thought it would be smaller package size for projects with a bit more more in runtime. The function is pretty small.

Overall for function handlers, for the most part everything stays the same, except when the handler props are passed in from the outside of the component - the delegation is done at runtime but the effect should be insignificant as the function call applies the same exact mechanism: attached `__click = handler` and calls `delegate` as before.

### DIFF
--- a/packages/ripple/src/compiler/phases/2-analyze/index.js
+++ b/packages/ripple/src/compiler/phases/2-analyze/index.js
@@ -2,7 +2,7 @@ import * as b from '../../../utils/builders.js';
 import { walk } from 'zimmerframe';
 import { create_scopes, ScopeRoot } from '../../scope.js';
 import {
-	get_delegated_event,
+	is_delegated_event,
 	get_parent_block_node,
 	is_element_dom_element,
 	is_inside_component,
@@ -729,16 +729,15 @@ const visitors = {
 						}
 
 						if (is_event_attribute(attr.name.name)) {
-							const event_name = attr.name.name.slice(2).toLowerCase();
 							const handler = visit(attr.value, state);
-							const delegated_event = get_delegated_event(event_name, handler, state);
+							const is_delegated = is_delegated_event(attr.name.name, handler, context);
 
-							if (delegated_event) {
+							if (is_delegated) {
 								if (attr.metadata === undefined) {
 									attr.metadata = {};
 								}
 
-								attr.metadata.delegated = delegated_event;
+								attr.metadata.delegated = is_delegated;
 							}
 						} else if (attr.value !== null) {
 							visit(attr.value, state);
--- a/packages/ripple/src/compiler/phases/3-transform/client/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/client/index.js
@@ -40,7 +40,11 @@ import {
 import is_reference from 'is-reference';
 import { object } from '../../../../utils/ast.js';
 import { render_stylesheets } from '../stylesheet.js';
-import { is_event_attribute } from '../../../../utils/events.js';
+import {
+	get_original_event_name,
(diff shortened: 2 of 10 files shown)


trueadm:
What about for attributes that start with `on` but aren't events? Like if there was `only` or something else?

leonidaz:
If we can’t statically analyze it, then it will be treated as an event.  Words that start with on are pretty uncommon, `only` being an exception.  That’s the limitation.

But we could keep the capitalization requirement, and either just require that custom events start with a capital letter, or if custom is true only the first capital letter will be lower-cased, or we also introduce a `customName` option that you can provide for the object notation and we’ll use that instead, and otherwise, we lowercase everything as before - and we drop the `custom` object option. 

I think the `customName` or it could be just `name` or `eventName` is the cleanest solution tbh.

leonidaz:
Actually, I'm now convinced that `customName` is the way to go. While keeping our 3rd letter capitalization rule to confidently identify events. `customName` (or whatever name we end up picking) allows all kinds of characters and with custom events it's typical to use a namespace as prefix and use some kind of a separator like colon `:`, e.g. `calculator:press`

leonidaz:
we could also keep the `Capture` suffix if we want it for DX / convenience with function handlers.  I'll just convert it to a `capture: true` object notation.

trueadm:
The `Capture` suffix is actually important for spreading. If you wanted to have both a bubbling `onClick` and a non-bubbling one in props, there would no other way to express that.

leonidaz:
> no other way to express that

yeah, not with a function handler, one would have to be an object. But yeah, I agree that we should just keep it. I'll add it back.

leonidaz:
ok, made the changes:
- went back to our original requirement to have a capital letter for the third position
- added `customName` to specify what the actual event name should be vs the attribute name
- added back `Capture` suffix
  - NOTE: for object notations if options has `capture: false` it will override the `Capture` suffix

trueadm (approved):

PR #562, opened 2025-11-13
update outdated example in docs

sreekar339339:


website/docs/guide/reactivity.md
```diff
@@ -218,35 +218,7 @@ function createDouble([ count ]) {
...
-    console.log('Count:', @count)
-  });
-
-  return { double };
-}
-
-export component App() {
-  let count = track(0);
-  const { double } = createDouble({ count });
+  const [ double ] = createDouble(count);
```
leonidaz: We we should drop the array as well, and just use the variable. Also is this also an issue in llms.txt?

leonidaz (approved):
thanks!

PR #565, opened 2025-11-14
fix(docs): correct invalid Html attribute in Attribute Binding example

marstrx:
The current Attribute Binding example in the Ripple documentation
uses <input type="textarea" /> which is not a valid HTML input type.
Replaced  <input type="textarea" />  with <input type="text" />.


leonidaz (approved):
Thanks!

marstrx:
my pleasure

PR #568, opened 2025-11-15
fix: don't include parser.c in linguist language graph

tijnjh:
the tree sitter plugin has a 4mb generated parser.c file

all this pr does is configure linguist (what github uses for the repo language graph) to ignore that specific file, to make the graph more realistic

before:
image

after:
image


leonidaz (approved):
Thank you!

Issue #341, opened 2025-09-27
[Bug]: the style tag doesn't have intellisense like other parts (html/js)


sina-salahshour:
### Issue Category

Compilation Bug

### Bug Description

when i hover over css attributes, nothing shows up.
also it doesn't have any autocomplete.

also i don't know if it's intentional but ripple files dont export their types and are typeless.

### Reproduction Steps

when i hover over css attributes, nothing shows up.
also it doesn't have any autocomplete.

### Environment

```markdown

```

### Checklist

- [x] I've searched for existing issues
- [ ] I'm willing to help implement a fix

PR #577, opened 2025-11-21
feat: add css service

kovsu:
fix: #341

https://github.com/user-attachments/assets/6be9b155-61f9-46b2-8ddf-487370d5a5b1

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,5 @@
   },
   "dependencies": {
     "@types/eslint": "catalog:default"
-  },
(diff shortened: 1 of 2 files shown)


leonidaz:
Thanks! definitely a great idea and appreciated! but we'll have to do it differently as far loading css since we already have it parsed out.  Plus i gotta take care of types, fix some integration bugs and make it coordinate with other plugins that we have in the works but not committed yet.

kovsu:
Completed by https://github.com/Ripple-TS/ripple/commit/87b2197d31d960cdd2e0c98501d68a896b379b1f. Thank you ❤️ @leonidaz.

PR #579, opened 2025-11-23
docs: correct pass text as expression inside braces in ripple template

anshiq:
Screenshot 2025-11-22 at 8 06 30 PM

Improve docs with correct passing of text as expression inside braces in Template Syntax section under Guide in official docs.


leonidaz (approved):
Thank you!

PR #580, opened 2025-11-23
fix: add templates to workspace and configure Vite for browser runtime

Dreezy305:
- Add templates/* to pnpm-workspace.yaml to enable workspace linking
- Update templates/basic/package.json to use workspace:* for local packages
- Configure Vite to prioritize browser conditions for proper runtime resolution
- Fix vite-plugin type declarations to use proper ES module exports


WebEferen:
Basic template should use latest instead of workspace, because we are using it with degit once we are scaffolding the application using `npx create ripple`. With such change we won't be able to resolve the dependencies version.

Please check how `create-app` works and how we are scaffolding the application.

Issue #576, opened 2025-11-20
[Bug]: Typescript files cannot handle Typescript


kuechlerm:
### Issue Category

Compilation Bug

### Bug Description

ts files in a ripple project have errors for basic ts things like
- `export type Test = { a:string};`
- `function a<T>(value: T): T { return value; }`

<img width="696" height="55" alt="Image" src="https://github.com/user-attachments/assets/c98ade2c-2b1d-4254-afaf-e68c0820a685" />

But both show no errors in a ripple file. 

### Reproduction Steps

1. npx create-ripple
2. add examples to a .ts file

### Environment

```markdown
- macOS
- VS Code
```

### Checklist

- [x] I've searched for existing issues
- [ ] I'm willing to help implement a fix

kovsu:
What's the error?

https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcAqwDO8C8cDecAhgFzZQCWAdgOYC+A3AFBMBmArlQMYwURXEAPGgB8ACgBuRADbtgJdAEoFaAnCjAY7KAKmzgDOHSA

kuechlerm:
@kovsu Your reproduction steps seem quite different from mine, no?

Added screenshot...

kuechlerm:
UPDATE: The same error occurs in nvim.

kovsu:
I mean it's a normal code in typescript and it doesn't make any error. Maybe you can check your eslint config or something.

kovsu:
now i can confirm it caused by eslint.

<img width="1128" height="306" alt="Image" src="https://github.com/user-attachments/assets/0e92390f-dbd6-4c74-9c51-8fa909fb0610" />

kuechlerm:
Ok, I figured out, that I needed to manually install "@typescript-eslint/parser"
Now it seems to work. That seems to be an error in the create-ripple setup or is it an pnpm thing?

PR #581, opened 2025-11-23
fix: eslint parsing error in template

kovsu:
fix: #576

Needs ts parser dependency for eslint in templates.

image


leonidaz (changes requested):
thanks!  actually, we should fix it by removing optional from both `@ripple-ts/eslint-parser` and `@typescript-eslint/parser`

here: packages/eslint-plugin/package.json, just remove the whole `peerDependenciesMeta` section

PR #583, opened 2025-11-23
feat: add platform-agnostic scripts for copying dependencies in vscode plugin

### LINKED ISSUE #500: [Feature]: Make all commands in all packages cross-platform
### Feature Category

Developer Experience

### Problem Statement

Currently many scripts defined in `package.json` don't run natively on Windows, which is not ideal. I can't use Linux, and I would strongly prefer not to use WSL just for running a few scripts, as it eats up about 1.5 GB of RAM just to have WSL running. Example of a script that's failing on Windows:

1. `cd packages/ripple-vscode-plugin`
2. `pnpm copy-ripple`

I realize that this is just a minor inconvenience, so it should be low priority, but it would be nice to have eventually.

### Proposed Solution

Replace platform-specific scripts with something cross-platform. Perhaps, we could write a few Node scripts that could replace those commands.

### Alternatives Considered

I considered:
* Using Linux, but I need to stay on Windows most of the time for Windows-specific things, and switching between them is not very convenient.
* Using WSL, but that is also not very convenient as it requires a lot of RAM to run.

### Examples & References

_No response_

### Checklist

- [x] I'm willing to help implement this feature

struckchure:
## Description

Adds cross-platform support for the `copy-ripple` script by detecting the platform and using appropriate commands for Windows vs Unix-based systems.

## Changes

- Added platform detection using `process.platform`
- Updated directory creation command to work on both Windows and Unix
- Replaced `rsync` with `robocopy` on Windows (since `rsync` isn't available by default)
- Maintained existing functionality on macOS/Linux

## Testing

- [x] Tested on macOS
- [ ] Tested on Windows
- [x] Verified that Ripple files are correctly copied to `dist/node_modules/ripple/`

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Fixes #500  (if applicable)

### DIFF
--- a/packages/vscode-plugin/scripts/copy-ripple.js
+++ b/packages/vscode-plugin/scripts/copy-ripple.js
@@ -6,13 +6,14 @@ const isWindows = process.platform === 'win32';
 
 try {
 	const createDirCommand = isWindows
-		? 'if not exist dist\node_modules (mkdir dist\node_modules)'
+		? 'if not exist dist\\node_modules mkdir dist\\node_modules'
 		: 'mkdir -p dist/node_modules';
-	const rsyncCommand =
-		'rsync -aL --del --exclude=node_modules/ripple --exclude=node_modules/typescript --exclude=.ignored* node_modules/ripple/ dist/node_modules/ripple/';
+	const copyCommand = isWindows
+		? 'robocopy node_modules\\ripple dist\\node_modules\\ripple /E /XD node_modules /XF .ignored* || (exit 0)'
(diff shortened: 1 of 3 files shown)


leonidaz:
please see this comment for reference: https://github.com/Ripple-TS/ripple/issues/500#issuecomment-3568369706

leonidaz:
I did most of the work yesterday because I had to and just added more today.

Thanks for jumping in but we needed js scripts for compatibility with all os's

PR #590, opened 2025-11-26
feat(language-server): add class-jump-to-style feature

wangshunnn:
## Feat (language-server)

Add class-jump-to-style feature to the Ripple IDE plugin (similar to Vue official plugin).
Support **static class** and **dynamic class** — clicking a class name in the template via `cmd + click`​ (or the link displayed on hover) jumps to the corresponding class name in the `<style>`, enhancing user experience.

## Implementation:

The core implementation is divided into the following stages:
1. Compile and parse class names in the template and their positions in the generated code — `scopedClasses` .
2. Compile and parse class names in the <style>tag and their positions in the generated code — `cssClasses`.
3. Add `documentLinksPlugin` for `language-server` to enable the jump functionality.

**Some additional details:** The class names and the `<style>` for jumping must be within the same component scope. Therefore, I used `node.css.hash` and the compilation traversal order to ensure that `scopedClassesand` and `cssClassesare` correctly matched.

## Effect

IDE screenshot:

class-jump-to-style

### DIFF
--- a/packages/language-server/src/documentLinksPlugin.js
+++ b/packages/language-server/src/documentLinksPlugin.js
@@ -1,0 +1,109 @@
+/** @import { LanguageServicePlugin, DocumentLink } from '@volar/language-server' */
+/** @import { RippleVirtualCode } from '@ripple-ts/typescript-plugin/src/language.js') */
+
+const { URI } = require('vscode-uri');
+
+const DEBUG = process.env.RIPPLE_DEBUG === 'true';
+
+/**
+ * @param {...unknown} args
+ */
+function log(...args) {
+	if (DEBUG) {
+		console.log('[Ripple Document Links]', ...args);
+	}
+}
+
+/**
+ * @returns {LanguageServicePlugin}
+ */
+function createDocumentLinksPlugin() {
+	return {
+		name: 'ripple-document-links',
+		capabilities: {
+			documentLinkProvider: {},
+		},
+		create(context) {
+			return {
+				async provideDocumentLinks(document) {
+					const uri = URI.parse(document.uri);
+					const decoded = context.decodeEmbeddedDocumentUri(uri);
+					if (!decoded) {
+						return;
+					}
+					const [sourceUri, virtualCodeId] = decoded;
+					if (virtualCodeId !== 'root') {
+						return;
+					}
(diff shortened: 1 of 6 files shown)


packages/language-server/src/documentLinksPlugin.js
```diff
@@ -0,0 +1,109 @@
...
+						for (const { text, offset } of cssClasses) {
+							const start = styleDocument.positionAt(offset);
+							const end = styleDocument.positionAt(offset + text.length);
+							const target =
+								styleDocumentUri +
+								`#L${start.line + 1},${start.character + 1}-L${end.line + 1},${end.character + 1}`;
+							if (!cssClassesMap.has(text)) {
+								cssClassesMap.set(text, []);
+							}
+							// @ts-ignore
```
WebEferen, line 76: ts-ignore is not needed here I think :)

leonidaz:
i rebased it first and i'll start applying the changes that I mentioned on discord.

packages/language-server/src/documentLinksPlugin.js
wangshunnn: Ideally, TS should correctly narrow the type scope, but it doesn't. I've encountered similar issues while writing TS code (need to manually add the Non-null assertion operator `!`), and there are similar problems elsewhere in the codebase, such as [// @ts-ignore](https://github.com/wangshunnn/ripple/blob/88cb4975b4b97b793fad0383dd161e5ab6d151cb/packages/ripple/src/compiler/phases/3-transform/client/index.js#L2995).
So I think it's more likely a problem with TS itself. Also, thanks for the review.

leonidaz:
@wangshunnn  thanks again for providing this!  I had to do it very differently though and I just worked on a clean branch - a lot of moving pieces there that we had to have.  I used most of your class attribute extractor function in segments.js and credited you and the pr in the commit.

closing this pr, the feature was implemented via commit 484ba2c3bb45f613cb0437e254f43609595c76cb

wangshunnn:
Thank you for your excellent work on this improvement. I also noticed that you renamed the virtual style `embeddedId` from `style_${index}` to `style_${hash}` — this makes more sense and unexpectedly aligns with my previous assumptions.

PR #594, opened 2025-12-01
fix: add TypeScript types for dynamic elements (#550)

### LINKED ISSUE #550: [Bug]: Dynamic element types don't allow any attributes
### Issue Category

Behavioral Bug

### Bug Description

Types for dynamic elements don't allow any attributes.

### Reproduction Steps

1. Create a component:
```jsx
export component App() {
	let div = track('div');

	<@div />
	<@div class="container" />
}
```
2. The language server complains:
    * For the first div:
      ```
      Type '{}' is not assignable to type 'never'. ts (2322)
      ```
    * For the second div:
      ```
      Type 'string' is not assignable to type 'never'. ts (2322)
      ```

### Environment

```markdown
- Ripple v0.2.167
```

### Additional Context

_No response_

### Checklist

Dreezy305:
This PR adds capitalized versions of HTML element types to the
IntrinsicElements interface in jsx-runtime.d.ts:
- Basic elements: Div, Span, P, H1-H6
- Form elements: Button, Input, Form
- Link/media: A, Img
- Semantic HTML5: Section, Article, Header, Footer, Nav, Main, Aside

Each capitalized element has the same type signature as its lowercase
counterpart, ensuring dynamic elements accept HTMLAttributes and provide
proper TypeScript autocomplete and type checking.

Fixes #550

### DIFF
--- a/packages/ripple/src/jsx-runtime.d.ts
+++ b/packages/ripple/src/jsx-runtime.d.ts
@@ -56,7 +56,7 @@ declare global {
 		type Element = void;
 
 		interface IntrinsicElements {
-			// HTML elements with basic attributes
+			// HTML elements with basic attributes (lowercase)
 			div: HTMLAttributes;
 			span: HTMLAttributes;
 			p: HTMLAttributes;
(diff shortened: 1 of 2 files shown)


leonidaz:
Thanks! 

Not sure if it’s the right fix as it allows capitalized elements. So, most likely we need to find a different solution. But it gives us the reason why this issue exists.

trueadm:
We should re-visit this at some point.

leonidaz:
We have a plan how to fix these via jsx code gen and no more capitalization, just need to get to it.

PR #604, opened 2025-12-21
optimize getElementName method from recursive to iterrative approach

anshiq:


leonidaz:
i'm not sure this is worth a change, typically it would recurse a couple of times, no big deal, but it's much more readable.

anshiq:
@leonidaz , if this algo won't be used for deep structures, then yes we can close PR withour merging..

PR #608, opened 2026-01-13
add lifecycle methods

anubra266:
Add `onMount` and `onDestroy`

### onMount

The `onMount` function runs once when the component is mounted. It can optionally return a cleanup function that will be called when the component is destroyed.

```jsx
import { onMount } from 'ripple';

export component App() {
  onMount(() => {
    console.log('Component mounted');

    // Optional: Return cleanup function
    return () => {
      console.log('Cleanup from onMount');
    };
  });

  <div>{'Content'}</div>
}
```

### onDestroy

The `onDestroy` function registers a cleanup callback that runs when the component is destroyed.

```jsx
import { onDestroy } from 'ripple';

export component App() {
  onDestroy(() => {
    console.log('Component destroyed');
  });

  <div>{'Content'}</div>
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,9 @@ export component App() {
 
 **[→ Transporting Reactivity Guide](https://www.ripplejs.com/docs/guide/reactivity#transporting-reactivity)**
 
-### Effects & Side Effects
+### Effects & Lifecycle
+
+**Effects** run when dependencies change:
 
 ```jsx
 import { effect, track } from 'ripple';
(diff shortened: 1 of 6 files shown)


leonidaz:
Not sure if we’d have the same issue as svelte had with onDestroy 

https://github.com/sveltejs/svelte/pull/15469

Still not sure how I feel about introducing the lifecycle methods. People may use them unnecessarily because they are used to them in other frameworks. We do have ref and effect.

trueadm:
Yeah I'm not sure either, maybe it's better to have them though as people will just get them from some broken third party lib otherwise.

Yusyuriv:
In my opinion, it's much, much easier for people to just use `onDestroy()` than to figure out how to do teardown using `effect()` correctly. I prefer simplicity in DX, so my vote is for having lifecycle methods built-in.

trueadm:
> In my opinion, it's much, much easier for people to just use `onDestroy()` than to figure out how to do teardown using `effect()` correctly. I prefer simplicity in DX, so my vote is for having lifecycle methods built-in.

The downside is when it comes to async. Which we haven't fully explored yet, but their timing is already a problem with Svelte 5's `onDestroy` and their experimental async mode.

trueadm:
Let's close this for now, and maybe re-visit in the future once we have a better understanding around async

Issue #610, opened 2026-01-16
[Bug]: Unterminated JSX error in component with nested code


anschl:
### Issue Category

Compilation Bug

### Bug Description

I found a combination of code and markup that throws off the parser:
```tsx
component ErrorComponent() {
	<div>
		<div>
			<div>
				const a = 1;
			</div>
			const a = 1;
		</div>
		const a = 1;
	</div>
}
```
```bash
error during build:
[vite-plugin] Unterminated JSX contents (11:1)
```

I looked around in the codebase and found [a component in this test](https://github.com/Ripple-TS/ripple/blob/cb0422a679fbf628fb0267bbab93fa0abcea012b/packages/ripple/tests/client/compiler/compiler.basic.test.ripple#L83) which also fails to compile using this variation:
```tsx
it('compiles lexical blocks correctly', () => {
	const code = `
	component App() {
		<div>
			const a = 1;
			<div>
				const b = 1;
			</div>
			<div>
				const b = 1;
			</div>
		</div>
		<div>
			const a = 2;
			<div>
				const b = 1;
			</div>
		</div>
	}
	`;
	expect(() => compile(code, 'test.ripple', { mode: 'client' })).not.toThrow();
});
```

### Reproduction Steps

1. Open your IDE
2. Create the above component
3. Squiggly line appears

### Environment

```markdown
Ripple 0.2.196
```

### Checklist

- [x] I've searched for existing issues
- [x] I'm willing to help implement a fix

leonidaz:
should be closed by #613

anschl:
@leonidaz  Tested it on main, looks good 👍
I would love to see a test for it though to prevent regression. It would also make refactoring easier.

PR #614, opened 2026-01-20
fix: correct parser handling of JSX tags misinterpreted as less than operator #610

jayasurya261:
### Description #610 
This PR fixes a parser bug where nested lexical declarations (e.g., const or let) inside markup caused an Unterminated JSX contents error.
#610

### The Issue
When the parser encounters a statement like const a = 1; inside a nested element, it enters "Script Mode." After finishing the statement, Acorn automatically reads ahead to the next token.

Because the parser was just in Script Mode, it interprets the next character (often the < of a closing </div> tag) as a Relational Operator (Less Than) instead of a JSX Tag Start. This desynchronizes the parser, causing it to miss the closing tag and eventually crash.

## Failing Example:
``` 
component ErrorComponent() {
  <div>
    <div>
      <div>const a = 1;</div> // Works (deepest level)
      const a = 1; // Works
    </div>
    const a = 1; // CRASHES HERE: The parser reads `</div>` as `<` (operator) + `/div>` (regex)
  </div>
} 
```
## The Fix
In parseTemplateBody, immediately after this.parseStatement(null) returns:

We reset the parser position (this.pos) back to this.start.

We explicitly call this.readToken(code) to force re-tokenization.

This ensures that the next token is evaluated using the correct context (checking for JSX tags via getTokenFromCode) rather than assuming it is a continuation of a JavaScript expression.

## Type of Change
[x] Bug fix (non-breaking change which fixes an issue)
[ ] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Screenshot of the solution 
Screenshot 2026-01-20 144704

## Solution as code
```
// Rewind the parser to the start of the current token 
// This is the token that was wrongly interpreted as "Less Than" < operator
this.pos = this.start;

				
// We pass the character code of the current position (the '<' character code, usually 60).
// This forces the parser to re-evaluate it, correctly seeing it as a JSX Tag now.
this.readToken(this.input.charCodeAt(this.pos));
```
## File path on local machine
D:\Developer\RippleJS\610Bug\rippleCompiler\ripple\packages\ripple\src\compiler\phases\1-parse\index.js

### DIFF
--- a/packages/ripple/src/compiler/phases/1-parse/index.js
+++ b/packages/ripple/src/compiler/phases/1-parse/index.js
@@ -2133,6 +2133,16 @@ function RipplePlugin(config) {
 					const node = this.parseStatement(null);
 					body.push(node);
 
+					// Rewind the parser to the start of the current token 
+					// This is the token that was wrongly interpreted as "Less Than" < operator
+					this.pos = this.start;
+
+				
+					// We pass the character code of the current position (the '<' character code, usually 60).
+					// This forces the parser to re-evaluate it, correctly seeing it as a JSX Tag now.
+					this.readToken(this.input.charCodeAt(this.pos));
+
+
 					// Ensure we're not in JSX context before recursing
 					// This is important when elements are parsed at statement level
 					if (this.curContext() === tstc.tc_expr) {


leonidaz:
this issue doesn't exist on the latest fixes via this pr #613

leonidaz:
closing as this is fixed on main.

PR #617, opened 2026-01-21
chore: unifiy compiler init and setup phases

trueadm:
In order to get hydration working, we need the DOM `child` and `sibling` calls to be in order of control flow, not just added at the top of a component block. So they need to be in the `init` phase.


leonidaz:
seems like it should be fine. 👍

PR #636, opened 2026-02-11
feat: enable early return in components

WebEferen:
Early return (`return;`) inside top-level `if` blocks in components, enabling guard-clause patterns.
The return is a compile-time marker - the compiler negates the return condition and  wraps the rest of the body in a reactive conditional block.

Example:
```javascript
export component App() {
    let a = track(false);
    let b = track(true);
    let c = track(true);

    if (@a) {
        <div>{'a true'}</div>
        <div />
        return;
    }

    if (@b) {
        <div>{'b true'}</div>

        if(@c) {
            <div>{'c'}</div>
            return;
        }

        return;
    }
}
```

Then the response would be: `b true` and `c`

### DIFF
--- a/packages/ripple/src/compiler/phases/2-analyze/index.js
+++ b/packages/ripple/src/compiler/phases/2-analyze/index.js
@@ -40,31 +40,6 @@ import { validate_nesting } from './validation.js';
 const valid_in_head = new Set(['title', 'base', 'link', 'meta', 'style', 'script', 'noscript']);
 
 /**
- * Checks if a node contains template content (elements, text, or control flow with templates)
- * @param {AST.Node} node
- * @returns {boolean}
- */
-function has_template_content(node) {
-	if (
-		node.type === 'Element' ||
-		node.type === 'Text' ||
-		node.type === 'Html' ||
-		node.type === 'TsxCompat'
-	) {
-		return true;
-	}
-	if (
-		node.type === 'IfStatement' ||
-		node.type === 'ForOfStatement' ||
-		node.type === 'TryStatement' ||
-		node.type === 'SwitchStatement'
-	) {
-		return node.metadata.has_template === true;
-	}
-	return false;
-}
-
-/**
  * @param {AnalysisContext['path']} path
  */
 function mark_control_flow_has_template(path) {
@@ -136,7 +111,7 @@ function mark_as_tracked(path) {
 
 /**
  * @param {AST.ReturnStatement} node
- * @returns
+ * @returns {AST.ReturnStatement}
  */
 function get_return_keyword_node(node) {
 	const return_keyword_length = 'return'.length;
@@ -879,7 +854,7 @@ const visitors = {
 			error_return_keyword(
 				node,
 				context,
-				'Return statements inside components cannot have a return value',
+				'Return statements inside components cannot have a return value.',
 			);
 		}
 
--- a/packages/ripple/tests/client/return.test.ripple
+++ b/packages/ripple/tests/client/return.test.ripple
@@ -2,6 +2,21 @@ import { describe, it, expect } from 'vitest';
 import { track, flushSync } from 'ripple';
 import { compile } from 'ripple/compiler';
 
+describe('returns in prohibited scopes', () => {
+	it('throws error when return is used in module scope', () => {
+		expect(
+			() => compile(`
+				return;
+				component App() {
+					<div>{'hello'}</div>
+				}
+			`, 'test.ripple', {
+				mode: 'client',
+			}),
+		).toThrowError('Return statements are not allowed at the top level of a module.');
+	});
+});
+
 describe('early return in client components', () => {
 	it('skips template content after direct return', () => {
 		component App() {
@@ -1962,7 +1977,7 @@ describe('early return in client components', () => {
 					}`, 'test.ripple', {
 					mode: 'client',
 				});
-			}).toThrow('Return statements inside components cannot have a return value');
+			}).toThrow('Return statements inside components cannot have a return value.');
 		});
 
 		it('throws error for return with number value', () => {
@@ -1975,7 +1990,7 @@ describe('early return in client components', () => {
 					}`, 'test.ripple', {
 					mode: 'client',
 				});
-			}).toThrow('Return statements inside components cannot have a return value');
+			}).toThrow('Return statements inside components cannot have a return value.');
 		});
 
 		it('throws error for return with null value', () => {
@@ -1988,7 +2003,7 @@ describe('early return in client components', () => {
 					}`, 'test.ripple', {
 					mode: 'client',
 				});
-			}).toThrow('Return statements inside components cannot have a return value');
+			}).toThrow('Return statements inside components cannot have a return value.');
 		});
 
 		it('throws error for return with undefined value', () => {
@@ -2003,7 +2018,7 @@ describe('early return in client components', () => {
 					'test.ripple',
 					{ mode: 'client' },
 				);
-			}).toThrow('Return statements inside components cannot have a return value');
+			}).toThrow('Return statements inside components cannot have a return value.');
 		});
 
 		it('throws error for return with object value', () => {
@@ -2018,7 +2033,7 @@ describe('early return in client components', () => {
 					'test.ripple',
 					{ mode: 'client' },
 				);
-			}).toThrow('Return statements inside components cannot have a return value');
+			}).toThrow('Return statements inside components cannot have a return value.');
 		});
 
 		it('throws error for return with variable', () => {
@@ -2034,7 +2049,7 @@ describe('early return in client components', () => {
 					'test.ripple',
 					{ mode: 'client' },
 				);
-			}).toThrow('Return statements inside components cannot have a return value');
+			}).toThrow('Return statements inside components cannot have a return value.');
 		});
 
 		it('accepts void return', () => {


leonidaz (changes requested):
this doesn't parse currently with this error message from acorn `'return' outside of function`

you need to enable `allowReturnOutsideFunction: true,` in the parser options: <https://github.com/Ripple-TS/ripple/blob/8ccf17996bb6fffaefa19760a7889101b5b11fb1/packages/ripple/src/compiler/phases/1-parse/index.js#L3167>

and after that, we need to add more checks in the analyzer so the return cannot appear anywhere else except inside components' if statements and functions.  E.g. after this change you could even return in the module.

We should also check that the inside component if's cannot return any type of value, just has to be `return` only.

So I was suggesting to support `return` even in nested scopes, nested elements, etc., like so:

```tsx
export component App() {
	<div id="first-top-block">{'Hello'}</div>

	<div id="second-top-block">
		<div>
			if (true) {
				return;
			}
		</div>
		<div id="sibling-block">{'Sibling'}</div>
	</div>

	<div id="third-top-block">{'Good bye'}</div>
}
```

so that calling a return there, would still finish the `#second-top-block` but would not render anything below. E.g. the `#third-top-block` would be wrapped in an if with a variable condition for each var that could be true. So, that's why the use of hoisted unique variables to enable this.

Also, in the nested return, the scopes below the return on the same level, or siblings, will not be rendered either. but any parent will be closed out. So, in this case, `#sibling-block` would be skipped.  I think it would really cool to support this as it seems much more valuable.

In the editor we'll show some kind of info / warning on the nested `return` to explain what to expect.  I think it totally makes sense that the scope will be closed out and siblings will be skipped.

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -2829,6 +2841,10 @@ function transform_ts_child(node, context) {
 		}
 		state.init.push(/** @type {AST.Statement} */ (/** @type {unknown} */ (result)));
 	} else if (node.type === 'ReturnStatement') {
+		// Skip early returns inside component if-with-return (handled by condition-negation wrapper)
+		if (!node.argument && is_inside_component(context) && !state.to_ts) {
+			return;
+		}
```
leonidaz: transform_ts_child is only called when `state.to_ts === true`

You don't need to do any checks here since the editor will show any errors from the analyzer (and at runtime, the errors will throw).

packages/ripple/src/compiler/phases/2-analyze/index.js
```diff
@@ -742,7 +789,11 @@ const visitors = {
 
 		context.visit(node.consequent, context.state);
 
-		if (!node.metadata.has_template) {
+		const consequent_body =
+			node.consequent.type === 'BlockStatement' ? node.consequent.body : [node.consequent];
+		check_unreachable_after_return(consequent_body, context);
```
WebEferen: If we don't want compiler error, we can simplify this even more (by removing the checks + helper method) @leonidaz

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -2913,7 +3031,70 @@ function transform_children(children, context) {
...
+					}
+					return initial;
+				} else if (state.flush_node !== null) {
+					const id = b.id(state.scope.generate('node'));
+					state.init?.push(b.var(id, b.call('_$_.child', state.flush_node?.())));
+					return id;
+				} else {
+					return b.id(state.scope.generate('node'));
+				}
+			})();
```
leonidaz: you shouldn't need to be doing anything with building nodes.

the idea is to identify children, wrap each in an `if` by create an `if` node via `b.if` and specify the child as the consequent, and the `test` as the unique variable condition (`if (__r_1) `), then visit this new `if` node instead of the child.  

Then, in the `if` visitor, set a new state with a flag for the ReturnStatement and let it add an assignment of the unique to true, instead of the the return: `__r_1 = true;` (or _$_.set if a tracked needed).

WebEferen: Done. Now we are not creating new nodes, we are reusing existing ones.

packages/ripple/src/compiler/phases/3-transform/client/index.js
WebEferen: Fixed.

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -2913,7 +3031,15 @@ function transform_children(children, context) {
...
 	};
 
-	for (const node of normalized) {
+	for (let node_idx = 0; node_idx < normalized.length; node_idx++) {
+		let node = normalized[node_idx];
+		const original_node = node;
+
+		if (accumulated_return_flags.length > 0 && is_template_or_control_flow(node) && !state.to_ts) {
+			const guard_condition = build_return_guard([...accumulated_return_flags]);
+			node = b.if(guard_condition, node, null);
```
leonidaz: actually, we don't wanna replace nodes and end up creating `if` and then visiting it unnecessarily which actually creates some issues.

We can collect the children by specifying a new state `const init = []; {...state, init}`, then once all the visitors hit, only then create the `if` wrapper with the `init[]` and we don't want to visit this `if` again, just add it to the original `state.init`. And if no wrapper needed, just state.init.push(...init)

This will also us to do an optimization to reduce the number of `if` wrappers.  The elements at the same level that don't have a return themselves can be wrapped into 1 `if` vs an `if` for each one.

leonidaz: Also, if the `if` only contains a return, we can just directly print variable = true vs creating a consequent and calling the variable = true inside.
for:
```jsx
		if (show && @b) {
			return;
		}
```
instead of this:
```jsx
		{
			var consequent = (__anchor) => {
				_$_.set(__r, true);
			};

			_$_.if(node, (__render) => {
				_$_.set(__r, false);

				if (show && _$_.get(b)) __render(consequent);
			});
		}
```
print this:
```jsx
		{
			_$_.if(node, (__render) => {
				_$_.set(__r, false);

				if (show && _$_.get(b)) _$_.set(__r, true);
			});
		}
```

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -3084,27 +3215,31 @@ function transform_children(children, context) {
 				node.is_controlled = is_controlled;
 				visit(node, {
 					...state,
+					return_flags,
```
leonidaz: do we need this here since for loops should not have returns or children that contain returns? same for other places a return should never occur.

packages/ripple/src/compiler/phases/2-analyze/index.js
```diff
@@ -752,19 +828,73 @@ const visitors = {
...
+			}
+
+			if (!ancestor.metadata.returns) {
+				ancestor.metadata.returns = [];
+			}
+			ancestor.metadata.returns.push(node);
+			ancestor.metadata.has_return = true;
 		}
+
+		node.metadata.is_reactive = is_reactive;
```
leonidaz: I think this can be just done in the `IfStatement` visitor, by setting a new state with a local metadata that has `tracking: false`, see how we're doing it in the transformation stage. And then checking if `metadata.tracking` changed to `true`.

We should probably set `node.tracked = true` on the `IfStatement`.

I think we're only missing a NewExpression visitor in the analyze phase, so it'd need to be added.

```jsx
NewExpression(node, context) {
    if (context.state.metadata?.tracking === false) {
        context.state.metadata.tracking = true;
    }

    context.next();
},
```

So, then we don't need this function `has_tracked_in_expression`

packages/ripple/src/compiler/phases/3-transform/client/index.js
leonidaz: and you can determine if the `if` only has a lone `return` in the analyze phase.

packages/ripple/src/compiler/phases/2-analyze/index.js
leonidaz: > We should probably set `node.tracked = true` on the `IfStatement`.

either `if` of `test` expression node, test is more precise.

leonidaz (approved):
this lgtm

we perhaps can do something easier with the client transformer, but we can look into that later, i tried but not having written the code, it just would take too long.

@WebEferen great work!

@trueadm please review when you can.  This does allow us to return anywhere in the component and have if guards to skip rendering siblings and top level siblings if the return was nested somewhere in one of the top siblings.

i think this is a big win honestly!

leonidaz:
and, ssr and the editor support is also there, so we're good.

WebEferen:
> this lgtm
> 
> we perhaps can do something easier with the client transformer, but we can look into that later, i tried but not having written the code, it just would take too long.
> 
> @WebEferen great work!
> 
> @trueadm please review when you can. This does allow us to return anywhere in the component and have if guards to skip rendering siblings and top level siblings if the return was nested somewhere in one of the top siblings.
> 
> i think this is a big win honestly!

Love it! thanks for help and your refactors + edgecase finding :)
Hopefully we'd ship it shortly to enable waaay better features 🔥

trueadm:
Is hydration support in too? I'm happy to merge if so and if we have tests around it

trueadm:
We can hydration after.

PR #650, opened 2026-02-15
docs: add note about multiple components per file

balazshevesi:
I just added one line to the the comparison page in the docs


leonidaz (approved):
Thank you!

PR #694, opened 2026-02-18
feat: new website

trueadm:


packages/ripple/src/compiler/phases/2-analyze/index.js
```diff
@@ -144,6 +144,78 @@ function error_return_keyword(node, context, message) {
 	);
 }
 
+/**
+ * @param {AST.Expression} expression
+ * @returns {AST.Expression}
+ */
+function unwrap_template_expression(expression) {
```
leonidaz, line 151: @WebEferen 

do you mind getting this logic into a pr, so we can merge it into main?

I will do the same with my fixes.

packages/ripple/tests/client/basic/basic.errors.test.ripple
```diff
@@ -61,6 +61,34 @@ describe('basic client > errors', () => {
...
+	it('should throw error for interpolating props.children as text', () => {
+		const code = `
+			export component Layout(props) {
+				<div>{props.children}</div>
+			}
+		`;
+
+		expect(() => {
+			compile(code, 'test.ripple');
+		}).toThrow(
```
leonidaz, line 87: and sorry, @WebEferen when i fixed @children, didn't realize they were used in error messages, so these need to remove @

PR #698, opened 2026-02-19
fix: self-closing tag

WebEferen:
Fixed self-closing tag where nodes have properties and components are not dynamic.

### DIFF
--- a/packages/ripple/tests/server/basic.attributes.test.ripple
+++ b/packages/ripple/tests/server/basic.attributes.test.ripple
@@ -454,6 +454,6 @@ describe('basic server > attribute rendering', () => {
 		}
 
 		const { body } = await render(Basic);
-		expect(body).toBeHtml('<div data-disabled="" /><input disabled />');
+		expect(body).toBeHtml('<div data-disabled=""></div><input disabled />');
 	});
 });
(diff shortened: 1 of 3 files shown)


packages/ripple/tests/server/basic.attributes.test.ripple
```diff
@@ -454,6 +454,6 @@ describe('basic server > attribute rendering', () => {
 		}
 
 		const { body } = await render(Basic);
-		expect(body).toBeHtml('<div data-disabled="" /><input disabled />');
+		expect(body).toBeHtml('<div data-disabled=""></div><input disabled />');
```
leonidaz, line 457: Why did this change?

WebEferen: During the hydration phase, it was giving wrong elements alignment, because self-closing does not exists in HTML elements: https://developer.mozilla.org/en-US/docs/Glossary/Void_element#self-closing_tags

So actually here this is proper behavior (with param it should open it and close).
This is the server response which should be semantically valid (HTML).

leonidaz: Sorry, was on the phone and didn’t think of the context

trueadm (approved):

PR #701, opened 2026-02-19
chore: missing analyze changes

WebEferen:
Added missing analyze changes (unwrap).

### DIFF
--- a/packages/ripple/src/compiler/phases/2-analyze/index.js
+++ b/packages/ripple/src/compiler/phases/2-analyze/index.js
@@ -1298,7 +1298,7 @@ const visitors = {
 			)
 		) {
 			error(
-				'`children` cannot be rendered using text interpolation. Use `<@children />` instead.',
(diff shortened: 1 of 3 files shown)


packages/ripple/src/compiler/phases/2-analyze/index.js
```diff
@@ -1218,6 +1290,21 @@ const visitors = {
...
 		mark_control_flow_has_template(context.path);
+
+		if (
+			is_children_template_expression(
+				/** @type {AST.Expression} */ (node.expression),
+				context.state,
+			)
+		) {
+			error(
+				'`children` cannot be rendered using text interpolation. Use `<@children />` instead.',
```
leonidaz: you need to use just `<children />`

`@` is only for dynamic components, e.g. via trackSplit

and change the errors messages in tests to reflect

WebEferen: True, right!

PR #718, opened 2026-02-20
make vite build work

trueadm:


leonidaz:
covered and combined in #721

PR #722, opened 2026-02-21
fix: switch inside for inside SVG crashing with undefined getter

anubra266:
Reproduction [here](https://www.ripple-ts.com/playground?v=0.2.213&title=Counter#config=code%2FN4IgLglmA2CmIC4QGED2BXAdmWAnEANCACawDOAxrhAA6SqaIiEgAWsAhsUwDwC2sMBwAEFVh1xlBAXgA6IAKoAVAGIBaABzzhAegB8szP0EjMHAXJAA3CLADuNVLjDaKDHNkt2IxMK2mkNhSwat6%2BrATCEJhQEBzQapTxsNIAjAB0AAza%2BiysYHzQAIJgYJJM0ByYAOaWsJiulWRklvIsQtVkiADaALpEHBSQVrAAosRQTkyU1HQsfBIA1ug0iKCVNegc1fBI%2BYUsbtj1YLwztGDCYACeNCnyfKjE6HDyBg2QfI7OwsDCj1hLgBfYQAM1wqD4wgA5LMaHBoQBuQyyT7fS4AKmEHDIwnOdDBEKh8nSOnxYHScNeIGRDUwwlEDDIl1gAA90bjpMIAPIAIwAVrAhulFrBrmQABTkgCUtIZR2ZwiKNBowi5bI56TgNT8aukXNSKJcYAA-HiqBduhqnGAyN1Mr1ekbUQhhFKLXR0qRQRwXpcAD7%2B82zCnK1WB4OW63OMjpUHRYgS2BqvRXW6wVCgyN0K29PVc6GgrBDCAMaHCaW9WWGBkQLMSm53TNKlXCACE%2BphRcwJbL0t%2BNYZV1YELswkw9mEo1wENwEuhADlUMIrPEfMJo5dQRhMMRodX6cIgYOAdgJWHIsBBwyhLgdmBXcRUBR0AJsOleU9rukOCr6sRkFYCBoETJ8XzfCkqE4HBRjgCD5wmKx92lAhByBA8eDJD0wD0EAgSIZlrjgNYQA2aoth2JgKGaQ53BOJg8h8UhMDQY5sCYT9iGuX4wXcNQfT4YDrldNRf3hEIyHFHA%2BEiAAhaBokWABZQYAGUpNgPgVHcSJoVU2BqlQZMFAASWhSIACVUE-MBUEiblWWuHZMEiBReUBdBImQKpb1gaBoF07k7npVSqjIcyYQACT8kZIAoEQF1gdBYAisgwsSPA60RYReUGRZqghLBiFdVgyGgCVMkiTIAFJIlSDRqtlRloCcV0AGJQU6o88II7CSLIijdhAKl4CII4PFOJAIC%2BG0eLKPKj0JSEYRGpEUQndlZu9X1oEuNwZonbBhDQQE8Alfsr0PBVLggI4l1INVhG6bp5BoDg-HkVCSGIYh5H6F6QDej7CHkH7fpAR11oZHgyCsap3iHXjcDda6no6SJ3rKMg82bW6GHu2ALuvIcyG8MAxDdDoicPRH5RxZNoSB1hoQQYnaeEHgmd%2BdIecxyQQX0Nnad5XBOEWOVaePGmj0HTDYfhwwpZ6kBCLgMh2EELoED6XqQy1nWQGosgAAVRakSbmCIGgIWCZonH1-pDfQZlIX00pok6EiRpIkZJFLRgkEydIACYQ9SABmPD8JAaaOTWaPG3IeOiBwZktfWKpyO2IbE-JWi2It5XfbIf2mAAFgAdnaVBUGgdOQHqDheTgbgkHiaAWEGYYhoVWvRpVoQwGdphUGCqOgA)


packages/ripple/src/runtime/internal/client/template.js
```diff
@@ -84,6 +88,15 @@ export function template(content, flags) {
 	return () => {
 		if (hydrating) {
 			if (is_fragment) {
+				// Lazily create SVG/MathML fragments on first use — by this point
+				// init_operations() has been called by hydrate().
+				if (node === undefined) {
```
trueadm: We shouldn't be creating any nodes during hydration. This is not right.

anubra266: @trueadm would using native firstchild here instead be acceptable?
issue is `get_first_child` is populated by`init_operations`, but that happens after module level evaluation, i.e. after we lazy load the svg/mathml namespace.
So at the point we're currently trying to load the namespace, `first_child_getter` which we sue for `get_first_child` is still undefined

Image

trueadm: It shouldn't be undefined, `init_operations` should be called during the `hydrate` function being called.

trueadm: And this function should be called from within a component, not module level scope. Can you show me the stack where this is being called from module level scope?

anubra266: @trueadm came up with a different fix, the issue was we were trying to wrap anchor comments `<!>` in the svg fragment, we were doing it right somewhere down the function, it's in the lazy loading that it was wrong

packages/ripple/src/runtime/internal/client/template.js
```diff
@@ -76,8 +76,8 @@ export function template(content, flags) {
 	if (is_fragment) {
 		node = create_fragment_from_html(
 			has_start ? content : '<!>' + content,
-			use_svg_namespace,
-			use_mathml_namespace,
+			!is_comment && use_svg_namespace,
```
leonidaz: Yep, this the right fix. It’s my bad, should’ve included this known condition - too many refactors

leonidaz: actually, we should add !is_comment condition when we define `use_svg_namespace` and `use_mathml_namespace` because we define svg and mathml with it below and then we compare them to recreate the node if they differ.

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -3129,7 +3129,11 @@ function transform_children(children, context) {
 					(node.id.type !== 'Identifier' || !is_element_dom_element(node))),
 		) ||
 		normalized.filter(
-			(node) => node.type !== 'VariableDeclaration' && node.type !== 'EmptyStatement',
+			(node) =>
+				node.type !== 'VariableDeclaration' &&
+				node.type !== 'EmptyStatement' &&
+				node.type !== 'BreakStatement' &&
+				node.type !== 'ContinueStatement',
```
leonidaz, line 3136: i'm now wondering if this is an exhaustive list that we need here?

like FunctionDeclaration, Component, ReturnStatement, ThrowStatement, DebuggerStatement, ClassDeclaration, etc.

leonidaz: my only concern is now here.  I guess we need lots of tests.  @trueadm what do you think?

trueadm: Yes we need tests for each case here

trueadm (approved):
Let's merge to fix this, and follow up with some more test coverage

PR #732, opened 2026-02-22
fix: hydration switch block flattening

WebEferen:
Fixed switch block flattening issue. Now it supports each variant of switch.

### DIFF
--- a/packages/ripple/src/compiler/phases/3-transform/client/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/client/index.js
@@ -45,7 +45,6 @@ import {
 	is_inside_call_expression,
 	is_value_static,
 	is_void_element,
-	is_component_level_function,
 	is_element_dom_element,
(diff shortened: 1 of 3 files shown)


packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -3627,6 +3630,42 @@ function transform_children(children, context) {
...
+	const result = [];
+	for (const node of consequent) {
+		if (node.type === 'BlockStatement') {
+			result.push(.../** @type {AST.BlockStatement} */ (node).body);
+		} else {
+			result.push(node);
+		}
+	}
+	return result;
+}
```
leonidaz: do you mind putting this function into compiler utils.js and using it from both client and server so we have one copy of it.

leonidaz:
i found an issue with tsx generation with switch statements that are blocks, i'll deal with it separately.

PR #755, opened 2026-02-25
fix: improve update handling in flush_updates and add test for keyed for statements

### LINKED ISSUE #754: [Bug]: Keyed `for` loop over derived value prevents sibling reactive nodes from updating
When a keyed `for` loop iterates over a derived (computed) value, sibling reactive text nodes stop updating after state changes. The same code works correctly without the key clause.

anubra266:
Fixes #754

---

> [!NOTE]
> **Medium Risk**
> Touches core client runtime update flushing logic, which can affect render/update ordering across many components; the change is small but in a critical path and relies on correct traversal semantics.
> 
> **Overview**
> Fixes nested update traversal in `flush_updates` by tracking a stack of prior `containing_update` blocks, so leaving a nested `CONTAINS_UPDATE` scope restores the previous scope instead of clearing it.
> 
> Adds a regression test covering keyed `for` blocks driven by a derived list where updates must correctly propagate to sibling DOM/text nodes (e.g., item list and count).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 102ed5ad581785447675e402ab6c3ec14bcf07ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.cursor/rules/project.mdc
+++ b/.cursor/rules/project.mdc
@@ -127,6 +127,8 @@ Source Code (.ripple) → Parse → Analyze → Transform → Output (JS + CSS)
   prop, let, const, function, component, for_pattern)
 - **Reactivity analysis:** Marks tracked expressions, derives tracking metadata
 - **CSS scoping:** Hash-based class names via `CSS_HASH_IDENTIFIER`
+- **`#style` analysis:** Validates usage context, collects referenced classes,
+  cross-checks against standalone CSS selectors
 - **Server block analysis:** Tracks exports from `#server` blocks
 
 ### Phase 3: Transform (`packages/ripple/src/compiler/phases/3-transform/`)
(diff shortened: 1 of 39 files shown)


packages/ripple/src/runtime/internal/client/runtime.js
```diff
@@ -609,12 +609,15 @@ function flush_updates(root_block) {
 	var current = root_block;
 	var containing_update = null;
 	var effects = [];
+	/** @type {(Block | null)[]} */
+	var containing_update_stack = [];
```
trueadm: I think a linked list might be better here rather than using an array, as we're only using it as a one-way stack.

trueadm:
@cursor review PR

PR #760, opened 2026-02-26
add support for <style /> treated as normal dom element through `raw` attrobute

anubra266:
> [!NOTE]
> **Medium Risk**
> Touches core compiler parsing/transform paths for `style` handling; mistakes could change CSS emission/scoping or SSR output ordering, but changes are narrowly guarded by the new `raw` attribute and `styleScopeHash` checks.
> 
> **Overview**
> Adds support for `<style raw>` so it is **not treated as component-scoped CSS** and instead compiles as a normal `<style>` DOM element.
> 
> Parsing now detects the `raw` attribute and skips attaching the stylesheet to `component.css`/`styleScopeHash`; client/server transforms and source-region extraction gate scoping behavior on `metadata.styleScopeHash`, including emitting raw styles inline (server) and keeping them in templates rather than stripping them during normalization. New client and server tests cover raw-only and mixed raw+scoped `<style>` behavior, ordering, and portal/iframe scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7d81dad4f3589921efac0b53c9a997d82641f1f7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ripple/src/compiler/phases/3-transform/segments.js
+++ b/packages/ripple/src/compiler/phases/3-transform/segments.js
@@ -89,7 +89,12 @@ function visit_source_ast(ast, src_line_offsets, { regions, css_element_info }) 
 	walk(ast, null, {
 		Element(node, context) {
 			// Check if this is a style element with CSS content
-			if (node.id?.type === 'Identifier' && node.id?.name === 'style' && node.css && node.metadata.styleScopeHash) {
+			if (
+				node.id?.type === 'Identifier' &&
+				node.id?.name === 'style' &&
+				node.css &&
+				node.metadata.styleScopeHash
+			) {
 				const openLoc = /** @type {ESTreeJSX.JSXOpeningElement & AST.NodeWithLocation} */ (
 					node.openingElement
 				).loc;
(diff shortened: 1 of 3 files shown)


trueadm:
So this attaches a style element to the head whilst rendering?

leonidaz:
I think I'd prefer if to use `type` or `lang` to indicate something different: `type="text/css"` or `type="css"` or `lang="css"` or ... tailwind

then we can allow any number of `<style type=whatever>` and nested ones that one can use. these will be inlined into dom (actually for `inlined` see below).

and in the future provide a preprocessor api.

For scoped `<style>` and actually for any other type maybe we also have a boolean attribute `inlined` where we place the style block where it's specified.  `<style inlined>`.  This will allow scoped classes for `<iframe>`.

trueadm:
Can someone explain why Ripple needs this? Why can't you just use global styles or `<style>` in the head?

leonidaz:
I think there is an issue with Portals and head and as I said last night, we should figure that out first and fix any bugs.

But we need to understand the limitations first as I was getting runtime errors and weird behavior with Abraham's example.

trueadm:
Let's aim to fix bugs before suggesting new features to get around them :P

PR #765, opened 2026-02-27
Ripple grammar termination patterns

leonidaz:
Complete the fix for missing `pending` keyword in shared termination patterns to ensure correct syntax highlighting for `var`/`const`/`using` statements.

---

> [!NOTE]
> **Low Risk**
> Low risk: updates TextMate grammar regex termination lists to include `pending`, affecting only syntax highlighting behavior.
> 
> **Overview**
> Fixes Ripple’s TextMate grammar termination regexes by adding `pending` to the shared keyword lists used to end `var`/`const` declarations and `using` statement patterns.
> 
> This prevents those constructs from incorrectly continuing past a `pending` control keyword, improving syntax highlighting correctness.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8d5edd70b16b75c9116ee25afb8ca3db9250dd49. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/grammars/textmate/ripple.tmLanguage.json
+++ b/grammars/textmate/ripple.tmLanguage.json
@@ -450,7 +450,7 @@
         {
           "name": "meta.var.expr.js",
           "begin": "(?=(?<![_$[:alnum:]])(?:(?<=\\.\\.\\.)|(?<!\\.))(?:(\\bexport)\\s+)?(?:(\\bdeclare)\\s+)?\\b(var|let)(?![_$[:alnum:]])(?:(?=\\.\\.\\.)|(?!\\.)))",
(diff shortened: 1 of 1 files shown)


leonidaz:
@cursor

given that pending can only be a part of try as in try/pending where catch and finally are optional and the fact that try/pending can only appear inside components (as opposed to functions, module level, and all others), is this fix still valid.

leonidaz:
@cursor review

given that pending can only be a part of try as in try/pending where catch and finally are optional and the fact that try/pending can only appear inside components (as opposed to functions, module level, and all others), is this fix still valid.

PR #790, opened 2026-03-07
fix(runtime): use block instead of stale b_block in reconcile_by_key Map path

### LINKED ISSUE #789: bug(runtime): update_value uses stale b_block instead of block in reconcile_by_key Map path
## Description

In `reconcile_by_key` (the keyed `for...of` reconciliation in `packages/ripple/src/runtime/internal/client/for.js`), the Map-based path (triggered for 32+ items) calls `update_value` with the wrong variable.

## Affected Code

**File:** `packages/ripple/src/runtime/internal/client/for.js`  
**Line:** ~453

```js
block = b_blocks[j] = a_blocks[i];  // correct block is assigned here
b_val = b[j];
if (is_indexed) {
    update_index(block, j);
}
update_value(b_block, b_val);  // BUG: b_block is stale/undefined
```

`b_block` is declared at line ~311 (`var b_block;`) and is **only assigned** inside the prefix-matching loop (lines ~316) and suffix-matching loop (~338). When those loops do not execute (e.g. a full list reversal where no prefix or suffix items match), `b_block` remains `undefined`. The Map-branch correctly assigns the matched block to `block` on line 448, but then mistakenly uses `b_block` for the value update.

## Fix

```js
update_value(block, b_val);
```

## Trigger Condition

Keyed `for...of` with **32+ items** where **no prefix or suffix items match** (e.g. full reversal of all items). Any key found in the Map lookup hits this crash.

`update_value` calls `set(block.s.v, value)`, which throws `TypeError: Cannot read properties of undefined (reading 's')` when `block` is `undefined`.

## Reproduction

```ripple
import { track } from 'ripple';

component KeyedFullReversal() {
  let items = track(
    Array.from({ length: 40 }, (_, i) => ({ id: i, text: `Item ${i}` }))
  );

  <div>
    for (let item of @items; index idx; key item.id) {
      <div>{idx + ': ' + item.text}</div>
    }
  </div>

  <button onClick={() => {
    @items = @items.toReversed();
  }}>
    {'Trigger crash'}
  </button>
}

export default KeyedFullReversal;
```

Click the button → `TypeError: Cannot read properties of undefined (reading 's')` inside `update_value`.

anubra266:
## Summary

- Fixes `update_value(b_block, b_val)` → `update_value(block, b_val)` in `reconcile_by_key`'s Map-based path
- `b_block` is only assigned in the prefix/suffix matching loops; when no prefix/suffix match exists, `b_block` remains `undefined`
- `block` is correctly assigned on the preceding line as the matched block from `a_blocks[i]`
- Without this fix, a full reversal of a keyed list with 32+ items causes `TypeError` inside `update_value`

Closes #789

---

> [!NOTE]
> **Low Risk**
> Small, targeted runtime fix plus a regression test; behavior change is limited to keyed list reconciliation during large reorders.
> 
> **Overview**
> Fixes keyed `for` reconciliation in `packages/ripple/src/runtime/internal/client/for.js` so the Map-based diff path updates the matched `block` (from `a_blocks[i]`) instead of a stale `b_block`, preventing crashes and ensuring values are updated during large keyed reorders.
> 
> Adds a regression test covering a full reversal of a 40-item keyed list (32+ items triggers the Map path) to assert DOM text updates correctly after reordering.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bdba12db53f1141bd0485a16cecdfc2b1f061fd3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ripple/tests/client/for.test.ripple
+++ b/packages/ripple/tests/client/for.test.ripple
@@ -175,9 +175,7 @@ describe('for statements', () => {
 
 	it('keyed for with 32+ items: full reversal updates values via Map path', () => {
 		component App() {
-			let items = track(
-				Array.from({ length: 40 }, (_, i) => ({ id: i, text: `Item ${i}` })),
-			);
+			let items = track(Array.from({ length: 40 }, (_, i) => ({ id: i, text: `Item ${i}` })));
 
 			<div>
 				for (let item of @items; index idx; key item.id) {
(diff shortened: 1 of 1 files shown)


trueadm:
Could use a test case to capture this

PR #814, opened 2026-03-26
fix(compiler): strip TypeScript class syntax from JS output

RazinShafayet2007:
Fixes #812
Fixes #813
## Summary
This PR fixes two related compiler bugs where TypeScript-only class syntax could leak into emitted JavaScript for `.ripple` files:
- `implements` clauses were preserved in JS output
- generic type arguments on class `extends` clauses were preserved in JS output
That produced invalid JavaScript such as:
```js
class PrintEvent implements BaseEvent {}
class StringMap extends Map<string, string> {}
```
and led to browser/build-time syntax errors.
## Root Cause
Ripple already strips many TypeScript-only constructs when compiling `.ripple` files to JavaScript, but class-level syntax was only partially covered.
Two gaps were involved:
1. In the class emit/transform path, `implements` clauses were not consistently removed in JS mode.
2. For `superclass` generics like `extends Map<string, string>`, the relevant AST metadata is carried on `superTypeParameters` for this parser shape, so handling only `superTypeArguments` was not enough.
As a result, valid TypeScript syntax inside `.ripple` files could survive into runtime JS output.
### Compiler output
Updated the client and server transform paths to strip TypeScript-only class syntax in JavaScript mode for both `ClassDeclaration` and `ClassExpression`, including:
- `implements`
- `typeParameters`
- `superTypeArguments`
- `superTypeParameters`
The client printer is also kept TypeScript-aware in `to_ts` mode, so editor/tooling-oriented TypeScript output continues to preserve the original syntax where appropriate.
### Regression coverage
Added regression tests for both client and server compilation to verify that:
- `class Foo implements Bar {}` emits valid JS without `implements`
- `class Foo extends Bar<string> {}` emits valid JS without generic type arguments on the `superclass`
### Files Changed
- `packages/ripple/src/compiler/phases/3-transform/client/index.js`
- `packages/ripple/src/compiler/phases/3-transform/server/index.js`
- `packages/ripple/tests/client/compiler/compiler.typescript.test.ripple`
- `packages/ripple/tests/server/compiler.test.ripple`
Snapshot files were updated accordingly.
#### `implements`
Source:
```ripple
class PrintEvent implements BaseEvent {}
```
Before:
```js
class PrintEvent implements BaseEvent {}
```
After:
```js
class PrintEvent {}
```
#### `extends type arguments`
Source:
```ripple
class StringMap extends Map<string, string> {}
```
Before:
```js
class StringMap extends Map<string, string> {}
```
After:
```js
[...]

PR #817, opened 2026-03-28
fix(compiler): preserve class `extends` generics in volar output

### LINKED ISSUE #816: [Bug]: `compile_to_volar_mappings()` crashes on classes with generic type arguments in `extends` clauses
### Issue Category

Compilation Bug

### Bug Description

While following up on #814, I found a related issue in Ripple’s TypeScript/tooling output path for class `extends` clauses with generic type arguments.
For code like:
```ripple
class StringMap extends Map<string, string> {}
export component App() {}
```
the `to_ts` / `compile_to_volar_mappings()` path should preserve the `superclass` type arguments.
There are two related problems here:
1. On the latest published version `(ripple@0.3.3)`, this does not crash, but the generated `to_ts` output drops the `superclass` generic type arguments.
2. On local code after #814, this same area now crashes because the TS printer path was updated to read `context.state.to_ts`, but that printer context does not expose `state`.

### Reproduction Steps

Run the following:
```ripple
import { compile_to_volar_mappings } from 'ripple/compiler';
const result = compile_to_volar_mappings(
	`class StringMap extends Map<string, string> {}
export component App() {}`,
	'test.ripple',
	{ loose: true },
);
console.log(result);
```

### Standalone Repro
- https://github.com/RazinShafayet2007/ripple-to-ts-class-extends-repro
Run locally:
git clone https://github.com/RazinShafayet2007/ripple-to-ts-class-extends-repro.git
cd ripple-to-ts-class-extends-repro
pnpm install
pnpm check:to-ts

#### Latest published version (0.3.3)
The generated to_ts output drops the superclass type arguments:
```ts
class StringMap extends Map {}
export function App() {}
```
#### **Local** code after #814
This crashes with:
```
TypeError: Cannot read properties of undefined (reading 'to_ts')
```

### Expected Behavior
The TypeScript/tooling output path should:
- not crash
- preserve superclass generic type arguments in class extends clauses
So this:
```ripple
class StringMap extends Map<string, string> {}
export component App() {}
```
should produce output containing:
```ts
class StringMap extends Map<string, string> {}
export function App() {}
```

### Root Cause
For this class shape, the parser stores the `superclass` generic instantiation on `superTypeParameters`, not `superTypeArguments`.
The TypeScript printer path for class declarations/expressions was already incomplete because it only handled `superTypeArguments`, so the generic type arguments were dropped in `to_ts` output.
Then, while following up on #814, additional `context.state.to_ts` checks in `create_tsx_with_typescript_support(...)` caused a regression because that printer context does not provide `context.state`, leading to a runtime crash.

### Environment

```markdown
- OS: Ubuntu 22.04.5 LTS
- Browser: Google Chrome 142.0.7444.134
- Ripple: 0.3.3 (published) and local main after #814
- Node.js: v22.17.1
```

### Additional Context

This appears related in spirit to #785 / #786, which fixed missing generic type arguments in the `to_ts` path for `interface extends ...` clauses. This issue is the class equivalent of that problem, and was made more visible by the follow-up regression after #814.

### Checklist

RazinShafayet2007:
Fixes #816 
Follow-up to #814

### Summary
This PR fixes the TypeScript/tooling output path for class `extends` clauses with generic type arguments.
For code like:
```ripple
class StringMap extends Map<string, string> {}
export component App() {}
```
the `to_ts` / `compile_to_volar_mappings()` output should preserve the `superclass` generic type arguments.
### Problem
There were two related issues in this area:
1. On the latest published version `(ripple@0.3.3)`, the `to_ts` output dropped `superclass` generic type arguments, so code like:
```ts
class StringMap extends Map<string, string> {}
```
was emitted as:
```ts
class StringMap extends Map {}
```
2. After #814, local code exposed a regression in the same TS printer path: additional `context.state.to_ts` checks inside `create_tsx_with_typescript_support(...)` caused a crash because that printer context does not provide `context.state`.
### Root Cause
For this class shape, the parser stores the `superclass` generic instantiation on `superTypeParameters`, not `superTypeArguments`.
The TypeScript printer for class declarations/expressions was not preserving `superTypeParameters`, which caused the generic type arguments to be dropped in `to_ts` output. The follow-up `context.state.to_ts` guard then turned this area into a runtime crash in the TS printer path.
### Fix
This PR updates the class printer inside `create_tsx_with_typescript_support(...)` so that it:
- preserves class `typeParameters`
- preserves class `implements`
- preserves `superclass` generic type arguments from `superTypeParameters`
- falls back to `superTypeArguments` only if needed
- avoids relying on `context.state.to_ts` inside the TS printer path

### Before / After
Using the output from my repro [ripple-to-ts-class-extends-repro](https://github.com/RazinShafayet2007/ripple-to-ts-class-extends-repro), `compile_to_volar_mappings()` previously dropped the `superclass` generic type arguments in the generated TypeScript output. With this fix, the same repro preserves them correctly.

#### Before:
```ts
class StringMap extends Map {}

export function App() {}
```
#### After:
```ts
class StringMap extends Map<string, string> {}

export function App() {}
```

### Checklist
- [x] Added a changeset
- [x] Added a regression test
- [x] Ran the targeted test suite
- [x] Ran the full test suite with `pnpm test` and formatting with `pnpm format`

---

> [!NOTE]
> **Medium Risk**
> Touches the TypeScript/TSX printing path used for Volar mappings; incorrect AST field handling could still drop type info or affect emitted TypeScript, but the change is narrow and covered by a regression test.
> 
> **Overview**
> Fixes Volar/`compile_to_volar_mappings()` TypeScript emission for classes so `extends` generic type arguments are preserved by printing `superTypeParameters` (and by removing TS-printer guards that relied on `context.state`).
> 
> Updates the AST typings to reflect `superTypeParameters` on class nodes, and adds a regression test asserting `class Foo extends Map<string, string>` remains intact in the generated Volar output.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 97cf0c3166022ca134443d0808f34160f82d3f84. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ripple/src/compiler/phases/3-transform/client/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/client/index.js
@@ -138,7 +138,6 @@ function visit_function(node, context) {
 function strip_class_typescript_syntax(node, context) {
 	delete node.typeParameters;
 	delete node.superTypeParameters;
-	delete node.superTypeArguments;
 	delete node.implements;
 
 	if (node.superClass?.type === 'TSInstantiationExpression') {
--- a/packages/ripple/src/compiler/phases/3-transform/server/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/server/index.js
@@ -81,7 +81,6 @@ function build_return_guard(flags) {
 function strip_class_typescript_syntax(node, context) {
 	delete node.typeParameters;
 	delete node.superTypeParameters;
-	delete node.superTypeArguments;
 	delete node.implements;
 
 	if (node.superClass?.type === 'TSInstantiationExpression') {
--- a/packages/ripple/src/compiler/types/index.d.ts
+++ b/packages/ripple/src/compiler/types/index.d.ts
@@ -104,14 +104,12 @@ declare module 'estree' {
 	interface ClassDeclaration {
 		typeParameters?: AST.TSTypeParameterDeclaration;
 		superTypeParameters?: AST.TSTypeParameterInstantiation;
-		superTypeArguments?: AST.TSTypeParameterInstantiation;
 		implements?: AST.TSClassImplements[];
 	}
 
 	interface ClassExpression {
 		typeParameters?: AST.TSTypeParameterDeclaration;
 		superTypeParameters?: AST.TSTypeParameterInstantiation;
-		superTypeArguments?: AST.TSTypeParameterInstantiation;
 		implements?: AST.TSClassImplements[];
 	}
 


packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -4971,19 +4971,19 @@ function create_tsx_with_typescript_support(comments) {
...
 			}
-			if (context.state.to_ts && node.typeParameters) {
+			if (node.typeParameters) {
 				context.visit(node.typeParameters);
 			}
 			if (node.superClass) {
 				context.write(' extends ');
 				context.visit(node.superClass);
-				if (context.state.to_ts && /** @type {any} */ (node).superTypeParameters) {
+				if (/** @type {any} */ (node).superTypeParameters) {
```
leonidaz: we shouldn’t be using type any. You can adjust the types in the compiler types. You’ll see examples what we do for other nodes. And then we don’t need to use if else and then just use super type parameters. More below.

RazinShafayet2007: Makes sense. Proceeding with the changes.

RazinShafayet2007:
@leonidaz I added `superTypeParameters` to the compiler types, removed the `any` casts in both the TS printer and the JS transform paths (in both client and server), removed the fallback to `superTypeArguments` and simplified the TS printer to use `superTypeParameters` directly. I also reran the targeted tests and local repros and everything passed.

packages/ripple/src/compiler/types/index.d.ts
```diff
@@ -103,12 +103,14 @@ declare module 'estree' {
 
 	interface ClassDeclaration {
 		typeParameters?: AST.TSTypeParameterDeclaration;
+		superTypeParameters?: AST.TSTypeParameterInstantiation;
 		superTypeArguments?: AST.TSTypeParameterInstantiation;
```
leonidaz: do you mind removing `superTypeArguments` from type definitions since they're not being used.

packages/ripple/src/compiler/types/index.d.ts
```diff
@@ -103,12 +103,14 @@ declare module 'estree' {
...
 		implements?: AST.TSClassImplements[];
 	}
 
 	interface ClassExpression {
 		typeParameters?: AST.TSTypeParameterDeclaration;
+		superTypeParameters?: AST.TSTypeParameterInstantiation;
 		superTypeArguments?: AST.TSTypeParameterInstantiation;
```
leonidaz: same here

packages/ripple/src/compiler/phases/3-transform/client/index.js
```diff
@@ -137,7 +137,7 @@ function visit_function(node, context) {
  */
 function strip_class_typescript_syntax(node, context) {
 	delete node.typeParameters;
-	delete (/** @type {any} */ (node).superTypeParameters);
+	delete node.superTypeParameters;
 	delete node.superTypeArguments;
```
leonidaz: then we don't need this line.

packages/ripple/src/compiler/phases/3-transform/server/index.js
```diff
@@ -80,7 +80,7 @@ function build_return_guard(flags) {
  */
 function strip_class_typescript_syntax(node, context) {
 	delete node.typeParameters;
-	delete (/** @type {any} */ (node).superTypeParameters);
+	delete node.superTypeParameters;
 	delete node.superTypeArguments;
```
leonidaz: this line can be deleted.

packages/ripple/src/compiler/types/index.d.ts
RazinShafayet2007: Sure. I was initially unsure whether `superTypeArguments` was still used anywhere else in the class AST path which is why I kept it. Since it's not being used, I’ll do the removal of it from the type definitions and the transform cleanup as well now.

leonidaz: if the parser uses superTypeParameters then superTypeArguments would not be used anywhere else.  you can double check with console.logs or the debugger (there are configs in launch.json - for vscode - the tsx generation, client and server generations)

RazinShafayet2007: Yeah, I confirmed that now. Also made the remaining cleanups.

leonidaz:
Thank you!

PR #819, opened 2026-03-30
fix(vite-plugin): invalidate virtual CSS module during HMR to prevent style desync

trueadm:
## Summary

Fixes #6973 — HMR update causes component styling to disappear.

## Root Cause

When editing a `.ripple` component's scoped CSS during development:

1. The **CSS hash changes** (it's derived from the `<style>` block content via `hash()`)
2. The Vite `transform` hook re-compiles the `.ripple` file, producing new JS with new hash-scoped class names (e.g., `class="ripple-NEWHASH"`) and new CSS with matching selectors
3. The `.ripple` module **self-accepts** via `import.meta.hot.accept()` — the HMR handler swaps the component function and the DOM re-renders with the **new** hash class names
4. **However**, the virtual CSS module (`?ripple&type=style&lang.css`) was **never invalidated** or included in the HMR update — the browser kept the **old** CSS with selectors targeting `.ripple-OLDHASH`
5. Result: **styles disappear** because the CSS selectors no longer match the DOM class names. Only a full dev server restart could recover.

## Fix

Inspired by [vite-plugin-svelte's approach](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/plugins/hot-update.js), the `hotUpdate` hook in `packages/vite-plugin/src/index.js` now:

1. **Uses `transformRequest`** instead of manually calling `compile()` — this runs the standard Vite pipeline (load → transform), which triggers the existing transform hook and updates `cssCache` as a side-effect, avoiding double-compilation
2. **Compares `cssCache` before/after** to detect actual CSS changes (skip no-op updates)
3. **Invalidates and includes the virtual CSS module** in the HMR update only when CSS actually changed, so the browser fetches fresh CSS in sync with the re-rendered component
4. **Uses the object form of `hotUpdate`** with `order: 'pre'` for proper timing

### Key differences from vite-plugin-svelte

- Svelte uses a separate `hot-update` plugin with a `post`-order transform to cache final output of all modules for before/after comparison. Ripple uses the existing `cssCache` for this comparison since it already tracks the CSS output per component.
- Svelte iterates all related modules and checks each one's transform result. Ripple takes a simpler approach: it only needs to check whether the CSS changed, since the `.ripple` module itself is already self-accepting for JS changes.

## Testing

- All existing tests pass (the 7 eslint-plugin failures are pre-existing due to `@ripple-ts/eslint-parser` package resolution, unrelated to this change)
- The HMR behavior is inherently a dev-server integration scenario that cannot be covered by unit tests
- Formatting checks pass

<div><a href="https://cursor.com/agents/bc-bfefb6fe-f84b-4ec5-9c22-d8768dd7bfff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bfefb6fe-f84b-4ec5-9c22-d8768dd7bfff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


trueadm:
@leonidaz what do you think of this PR?

PR #820, opened 2026-03-30
fix: sync `<select>` `bindValue` with typed and dynamic options

### LINKED ISSUE #810: [Bug]: bindValue on <select> doesn't pre-select the correct option
### Issue Category

Behavioral Bug

### Bug Description

`<select {ref bindValue(trackedVar)}>` doesn't pre-select the correct item in the select if `trackedVar` is pre-set. The select stays empty until the user makes another choice.

### Reproduction Steps

[Playground](https://www.ripple-ts.com/playground?v=0.3.3&title=Select#config=code%2FN4IgLglmA2CmIC4QGVZwMZhAGhAE1gGd0AnCAB0gHsA7REHEAC1gEM96AeAW1jFYAE6JqxKE%2BAXgA6IAKoAVAGIBaABwyBAegB8Umjz6CarXtJAA3CLADu5KiTAb0tMLBpgz1iHjBMJBS3RYZS8fJmwBCBooCFZoZWI42AkARgA6AAYNHUYmMG5oAEEwMDF6aFYaAHMzNycKwkIzGUZ%2BKsJEAG0AXVxWTAhzWABRPCh7emIySkZuUQBrAFdyRFAK6sXWKvgkPILGZ3c3LCROKYowATAAT3JkmW4qPEW4GV0aKUhuOwcBYAFHot3AIAL4CABmJCo3AEAHJpuQ4LCANx6T4Qb72S4AKgErEIAnOlAhUJhMjSmiJYDSCNeIFRHxoAiEtEIl1gAA8fmACRIBAB5ABGACtYJg0vNYNdCAAKKkASgZzMObIEhXI5AEfM53MIaTg1V8WokfJSaMcYAA-ITSBdOjqsYROhlut1zZ8EAI5bbKGkCODWC9LgAfYM26bU9Wa0Phu0Ohx68FRPAy2Ba7RXW6wKjg2OUe3dY182HgoEDWiwgTy7qKvTMiC5mU3O45tUagQAQhNcNLNHLNFh8r%2BdeZVyYUOsAhoNgEwxIUJIMthADkqAJzHFvAJ45dwVQgXhB0rQSPAe4ZVGIsAR8z%2BCRtmBPXgqOhFrx3GlBU9rmlWBq3HgADCTAQNAKbPq%2B77UqQbCuMMcBQUuYzmIO8rYCOIK1volI%2BmA2ggCCuBstccCrCA6xVJs2z0OgjQHC4xz0Lk3gEDQgEMe49Bfng1x-BCLjKAG3CgdcnrKH%2BiLBIQ0quNwEQAELQFE8wALL9MgMmwNwiguBEsKoFUVBprIACSsIRAASlQX5gFQET8hy1zbDQESyIKQJgIsESAZUd5oNAen8ncTLIJUhDmXCAASaBDJA6CCMusCLLAEWEGFCSwGQ4LIgIgr9PMVRQgenpMIQ0AyhkEQZAApBEKSqNViostA9iegAxOCnWggRRG4WRFFUTsIC0vAuCHK4nFIBi3J8YKyYAGpxMloIktCAgyCNMgMnoKqXM4LViFqAidCO-zeJ6KQRK4HKPutw2wHgGiEadkR4J6ABMV2crdMiCtAyVPehTJ8edAgAMxfTdnoyIVsB1CAoJA8yZ1vQIAAskM-SAhWsNcMgRBAhAACKE6wf0PZ6pTLc9NDdNt05cliLKYtOwLsQdqAYGAMpDtewNwJc4hcw97P2CZeBHW1I1pKU%2BUyu9WHMiOZxoGKlzACQsC5nNNB4It-2wHKquYCLVAHeL8ogu8o78SQXq7c19gCK2%2B32IQvM3qOnBUJQEC0OuS3JMArskGk3hgmMaXk3gEjB2b9hh8TpPR1bns23HB0y99IJpwInCaD71A0Nbo458D%2BdC2r7zK0wKTaMAAAGnNqxTAgACTAAAApXJtAfHJDiyCDcgvntfvCCPUgMRcCECwfAdAgPS9RGC9LyAtGEAACpr4gnAwuDkFCQSNG7XS9OvixstCqAlFE7RkSNZFDGIft0EgGRpGDn8EYRIDTY6qxf7NiIIA3Arg2QLzWJUSiWwhrAKpPRI4k0GC-2foQV%2B9A0YAHZWhUDNpAkAbgyZwA4EgOI0BGD9EgEMGirIzajSnvwTyC914tXEBwEEIIgA)

### Environment

```markdown
Ripple: 0.3.3
```

### Additional Context

_No response_

### Checklist

### LINKED ISSUE #818: [Bug]: `bindValue` on `<select>` does not reselect when the matching option is added later
### Issue Category

Behavioral Bug

### Bug Description

While working on #810, I tried to reproduce and explore more `bindValue` related edge cases and found another issue with `<select>`. `bindValue` on `<select>` does not reselect the correct option when the matching option is added later. If the tracked value is already set and the matching `<option>` is inserted asynchronously after mount, the `<select>` falls back to the first available option instead of selecting the newly added matching option.

### Reproduction Steps

[Playground](https://www.ripple-ts.com/playground?v=0.3.3&title=Counter#config=code%2FN4IgLglmA2CmIC4QGED2BXAdmWAnEANCACawDOAxrhAA6SqaIiEgAWsAhsUwDwC2sMBwAEFVh1xlBAXgA6IAKoAVAGIBaABzzhAegB8szP0EjMHAXJAA3CLADuNVLjDaKDHNkt2IxMK2mkNhSwat6%2BrATCEJhQEBzQapTxsNIAjAB0AAza%2BiysYHzQAIJgYJJM0ByYAOaWsJiulWRklvIsQtVkiADaALpEHBSQVrAAosRQTkyU1HQsfBIA1ug0iKCVNegc1fBI%2BYUsbtj1YLwztGDCYACeNCnyfKjE6HDyBg2QfI7OwsDCj1hLgBfYQAM1wqD4wgA5LMaHBoQBuQyyT7fS4AKmEHDIwnOdDBEKh8nSOnxYHScNeIGRDUwwlEDDIl1gAA90bjpMIAPIAIwAVrAhulFrBrmQABTkgCUtIZR2ZwiKNBowi5bI56TgNT8aukXNSKJcYAA-HiqBduhqnGAyN1Mr1ekbUQhhFKLXR0qRQRwXpcAD7%2B82zCnK1WB4OW63OMjpUHRYgS2BqvRXW6wVCgyN0K29PVc6GgrBDCAMaHCaW9WWGBkQLMSm53TNKlXCACE%2BphRcwJbL0t%2BNYZV1YELswkw9mEo1wENwEuhADlUMIrPEfMJo5dQRhMMRodX6cIgYOAdgJWHIsBBwyhLgdmBXcRUBR0AJsOleU9rukOCr6sRkFYCBoETJ8XzfCkqE4HBRjgCD5wmKx92lAhByBA8eDJD0wD0EAgSIZlrjgNYQA2aoth2JgKGaQ53BOJg8h8UhMDQY5sCYT9iGuX4wXcNQfT4YDrldNRf3hEIyHFHA%2BEiAAhaBokWABZQYAGUpNgPgVHcSJoVU2BqlQZMFAASWhSIACVUE-MBUEiblWWuHZMEiBReUBdBImQKpb1gaBoF07k7npVSqjIcyYQACT8kZIAoEQF1gdBYAisgwsSPA60RYReUGRZqghLBiFdVgyGgCVMkiTIAFJIlSDRqtlRloCcV0AGJQU6o88II7CSLIijdhAKl4CII4PFOJAIC%2BG0eN5BMADV4mSo9CUhGERqRQwFUueJoAAfVQOhS0wTlhG6Qc-h8V1oVSCKcFZB8YW5Cdy3wy6omKmEACZ7rZJ7oSUOxUDe1DDyur7oQAVj%2Bx6bpUCARlBwxelpTdGRmidsBbGgJX7K9Dx2vE-KFHBiDVYQ2pG9Iyjy%2BcYYPeUmUuI76FOimqdocS%2BjlSnqdgTrSYlPGUwHQ8mdOy5PgzdBLi5KQwCUaaZbAYX%2B2kVMCaHIdWZO2MaHQMhWAldJTb2w7jqZRmh3w4RUkyB3rYZXBBHQXB6RFjXRDgCQlYEDBVelgPrfQ3meCkOAhl%2BF2s3m3clugZKpRJoZYGIaUgXebXt1wN0id1hhhGbAvTuy0VuJL9IfHx68hx4EuV2WlJgErnxM5by3MBp-6gUwkus5twdMIj0mB54Gg9GAAADJRcDytPXQAEmAAABEfU%2BIIEp97nQJ8MIEepAQi4CN2BBC6BA%2Bl6kML6vkBqLIAAFF2FYYogaAhYJmicW-%2Bnvw3bJ8H0qUaInQSIjRIiMSQJ0mCZHSAAZgQXhfCIBpocjWCgxs5AMFEBwMyC%2B6wqjkW2ENLB5JaJsUmswFBUCyAwKQAAFgAOztFQKgaABCQD1A4LyOA3AkB7RYIMYYQ0FTsNGkfIQYBDZURalIbgQIgRAA)

### Behavior
After 1 second, the tracked value still remains `5`, but the `<select>` shows `One` as the selected option, even though the matching option should be `Five`. As a result, the model and DOM selection become out of sync. The expected behavior is that once the matching option is inserted, the `<select>` should automatically select `Five` so that the DOM stays consistent with the tracked value `5`.

### Environment

```markdown
- Ripple Version: [0.3.3]
```

### Additional Context

I found this while exploring additional `bindValue` related cases during work on #810. This seems related to `<select>` option mutation after mount: the tracked value is already correct, but when the matching option is inserted later, the selection is not reconciled against the existing bound value.

### Checklist

RazinShafayet2007:
Fixes #810 
Fixes #818 

## Summary
This PR fixes two related `<select {ref bindValue(...)}` bugs:
1. Typed (non-string) option values did not preselect correctly (`#810`)
2. A matching option added later did not get reselected automatically (`#818`)
Together, these could leave the tracked value and the visible DOM selection out of sync.
## Approach
This follows the same core approach Svelte uses for `<select>` bindings:
- preserve the raw bound value on `<option>` elements
- preserve the intended bound value on the `<select>`
- re-apply selection when option children/values change after mount
### Typed option values
The DOM always exposes `option.value` as a string. That meant a tracked value like `2` was compared against `"2"` and failed strict equality checks.
There was also a compiler-side gap: literal `<option value={...}>` could be emitted through the static attribute path instead of the runtime `set_value()` path, so the original typed value was never preserved.
### Late-added / changed options
When options were inserted or changed after mount, there was no mechanism to replay selection against the existing bound value. The browser fallback selection could win, causing the DOM and model to drift apart.
### Compiler
In `packages/ripple/src/compiler/phases/3-transform/client/index.js`:
- literal `<option value={...}>` is now excluded from the generic static-attribute fast path
- those values continue through the normal `value` handling path so runtime `set_value()` is used
This keeps non-`<option>` literal `value` attributes on their existing static/template path while ensuring `<option>` values get the runtime behavior they need.
### Runtime value preservation
In `packages/ripple/src/runtime/internal/client/render.js`:
- `set_value()` now stashes the original raw value on `<option>` elements via `__value`
In `packages/ripple/src/runtime/internal/client/bindings.js`:
- `get_option_value()` now prefers `option.__value` over `option.value`
This preserves typed option values such as numbers and objects.
### Dynamic option reconciliation
In `packages/ripple/src/runtime/internal/client/bindings.js`:
- the current intended selection is stored on `select.__value`
- a shared `MutationObserver` is used to observe relevant `<select>` elements
- when option children / `value` attributes change, selection is replayed with `select_option(select, select.__value)`
This fixes cases where matching options appear later or option lists are updated dynamically.
### Multiple Select Validation
Also in `bindings.js`, `<select multiple>` now throws a clear runtime error if the bound value is not an array:
```txt
Reactive bound value of a `<select multiple>` element should be an array, but it received a non-array value.
```
## Tests
Added regression coverage in:
- `packages/ripple/tests/client/input-value.test.ripple`
The added coverage includes:
- numeric option preselection
- getter/setter numeric option preselection
- preserving typed values on change
- reselection when a matching option is added later
- reselection when option lists are replaced later
- invalid non-array values on `<select multiple>`

All tests passed.

### DIFF
--- a/packages/ripple/src/runtime/internal/client/bindings.js
+++ b/packages/ripple/src/runtime/internal/client/bindings.js
@@ -333,14 +333,20 @@ export function bindValue(maybe_tracked, set_func = undefined) {
 
 				// Mounting and value undefined -> take selection from dom
 				if (mounting && value === undefined) {
-					/** @type {HTMLOptionElement | null} */
-					var selected_option = /** @type {HTMLOptionElement | null} */ (
-						select.querySelector(':checked')
-					);
-					if (selected_option !== null) {
-						value = get_option_value(selected_option);
+					if (select.multiple) {
+						value = [].map.call(select.querySelectorAll(':checked'), get_option_value);
 						select.__value = value;
 						setter(value);
+					} else {
+						/** @type {HTMLOptionElement | null} */
+						var selected_option = /** @type {HTMLOptionElement | null} */ (
+							select.querySelector(':checked')
+						);
+						if (selected_option !== null) {
+							value = get_option_value(selected_option);
+							select.__value = value;
+							setter(value);
+						}
 					}
 				}
 


RazinShafayet2007:
@trueadm @leonidaz have a look at this fix when you are free

packages/ripple/src/runtime/internal/client/bindings.js
leonidaz: would be great to also throw an error here.  at some point to centralize the error handling to reduce the bundle size.

'Reactive bound value of a `<select multiple>` element should be an array, but it received a non-array value.'

RazinShafayet2007: Implemented this and added regression tests for non-array values on `<select multiple>`.

packages/ripple/src/runtime/internal/client/bindings.js
```diff
@@ -192,6 +205,28 @@ function select_option(select, value, mounting = false) {
 	}
 }
 
+/**
+ * Re-applies the current bound selection when option children change after mount.
+ * @param {HTMLSelectElement & { __value?: unknown }} select
+ * @returns {() => void}
+ */
+function init_select(select) {
+	var observer = new MutationObserver(() => {
```
leonidaz: @RazinShafayet2007 just making sure you saw the discussion on #820 to create a shared MutationObserver with multiple `observe()` calls, disconnect, re-observe, etc.

I think everything else looks good and we can merge once that is added.

RazinShafayet2007: Yes, I saw the discussion. Was a little busy today but I'll start implementing the shared MutationObserver logic now.

RazinShafayet2007:
@leonidaz implemented the shared MutationObserver for selects and reran all the tests which passed. I also verified the behavior against my local repros to make sure the logic works in practice. Take a look at this and let me know if this approach looks good.

packages/ripple/src/runtime/internal/client/bindings.js
```diff
@@ -192,6 +205,87 @@ function select_option(select, value, mounting = false) {
...
+
+		if (current_select && observed_selects?.has(current_select)) {
+			selects.add(current_select);
+		}
+	}
+
+	for (const select of selects) {
+		select_option(select, select.__value);
+	}
+}
```
leonidaz, line 241: We can avoid running the second loop:

```suggestion
function process_select_mutation_entries(entries) {
	var selects = new Set();

	for (const entry of entries) {
      const target = /** @type {HTMLElement} */ (entry.target);
      const select = /** @type {HTMLSelectElement & { __value?: unknown } | null} */ (
        target.nodeName === 'SELECT' ? target : target.closest('select')
      );
		
        if (!select || selects.has(select)) {
          continue;
        }

		if (observed_selects?.has(current_select)) {
          selects.add(current_select);
          select_option(select, select.__value);
          continue;
		}
	}
}
```

packages/ripple/src/runtime/internal/client/bindings.js
```diff
@@ -192,6 +205,87 @@ function select_option(select, value, mounting = false) {
...
+
+/**
+ * @param {HTMLSelectElement & { __value?: unknown }} select
```
leonidaz: we should add this interface inside the `declare global` block here: packages/ripple/src/runtime/internal/client/types.d.ts

```ts
	interface HTMLSelectElement {
		__value?: unknown;
	}
```

and then we can just use `HTMLSelectElement`

We already do this for other Ripple-specific extensions.

RazinShafayet2007: done! made the changes.

PR #838, opened 2026-04-10
Fix TS types and add CI typecheck

trueadm:
> [!NOTE]
> **Medium Risk**
> Adds a new CI `typecheck` gate and adjusts compiler/runtime code to satisfy stricter TS checking, including a small behavior tweak in dynamic element detection that could affect compilation output.
> 
> **Overview**
> **Adds CI typechecking** by introducing a new `typecheck` GitHub Actions job and a root `pnpm typecheck` script that runs `tsc --noEmit` across multiple packages (with a new `packages/ripple/tsconfig.typecheck.json` scoped to runtime sources).
> 
> **Fixes TS/JSdoc typing issues** throughout the Ripple compiler and runtime (casts, `@ts-ignore` where needed, and more precise types in parser/runtime code), and updates `is_element_dynamic` to only treat tracked `Identifier`s (not `Literal`s) as dynamic. Also adds `@types/node` and updates the lockfile accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b48eaea6652a0a282842a680a3c9c6f9bba0e03f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/ripple/src/compiler/phases/1-parse/index.js
+++ b/packages/ripple/src/compiler/phases/1-parse/index.js
@@ -413,8 +413,9 @@ function RipplePlugin(config) {
 						/** @type {AST.Property} */ (prop).method = true;
 						/** @type {AST.Property} */ (prop).kind = 'init';
 						/** @type {AST.Property} */ (prop).value = this.parseMethod(false, false);
-						/** @type {AST.FunctionExpression} */ (/** @type {AST.Property} */ (prop).value).typeParameters =
-							typeParameters;
+						/** @type {AST.FunctionExpression} */ (
+							/** @type {AST.Property} */ (prop).value
+						).typeParameters = typeParameters;
 						return;
 					}
 				}
--- a/packages/ripple/src/compiler/phases/1-parse/index.js
+++ b/packages/ripple/src/compiler/phases/1-parse/index.js
@@ -413,7 +413,8 @@ function RipplePlugin(config) {
 						/** @type {AST.Property} */ (prop).method = true;
 						/** @type {AST.Property} */ (prop).kind = 'init';
 						/** @type {AST.Property} */ (prop).value = this.parseMethod(false, false);
-						/** @type {AST.Property} */ (prop).value.typeParameters = typeParameters;
+						/** @type {any} */ (/** @type {AST.Property} */ (prop).value).typeParameters =
+							typeParameters;
 						return;
 					}
 				}
@@ -705,7 +706,7 @@ function RipplePlugin(config) {
 						// & directly followed by { or [ — lazy destructuring
 						this.next(); // consume &, now current token is { or [
 						const pattern = super.parseBindingAtom();
-						pattern.lazy = true;
+						/** @type {AST.ObjectPattern | AST.ArrayPattern} */ (pattern).lazy = true;
 						return pattern;
 					}
 				}
@@ -898,6 +899,7 @@ function RipplePlugin(config) {
 					return this.parseFor(node, null);
 				}
 
+				// @ts-ignore — acorn internal: isLet accepts 0 args at runtime
 				let isLet = this.isLet();
 				if (this.type === tt._var || this.type === tt._const || isLet) {
 					let init = /** @type {AST.VariableDeclaration} */ (this.startNode()),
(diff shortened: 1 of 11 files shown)


packages/ripple/src/runtime/internal/server/index.js
```diff
@@ -1,6 +1,6 @@
 /**
 @import { Component, Dependency, Derived, Tracked } from '#server';
-@import { SSRComponent, renderToStream, render } from 'ripple/server';
```
leonidaz, line 3: let's keep this in place as I'm making changes to this anyway, but the types should work unless ts go does something different.

packages/ripple/src/runtime/internal/server/index.js
```diff
@@ -231,12 +231,12 @@ class Output {
 	}
 }
 
-/** @type {render} */
+/** @type {import('ripple/server').render} */
```
leonidaz, line 234: keep as before

packages/ripple/src/runtime/internal/server/index.js
```diff
@@ -262,7 +262,7 @@ export async function render(component) {
 	return { head, body, css };
 }
 
-/** @type {renderToStream} */
+/** @type {import('ripple/server').renderToStream} */
```
leonidaz, line 265: keep as before

packages/ripple/src/runtime/internal/client/head.js
```diff
@@ -54,7 +53,7 @@ export function head(hash, render_fn) {
 	} finally {
 		if (was_hydrating) {
 			set_hydrating(true);
-			set_hydrate_node(/** @type {TemplateNode} */ (previous_hydrate_node));
+			set_hydrate_node(previous_hydrate_node);
```
leonidaz, line 56: i can't comment on that line, but seems we need a type on the declaration of `previous_hydrat_node`: 
`let previous_hydrate_node = null;`

packages/ripple/src/runtime/internal/client/composite.js
```diff
@@ -29,6 +29,7 @@ export function composite(get_component, node, props) {
 
 	render(
 		() => {
+			// @ts-ignore — get() handles non-tracked values via is_ripple_object() check
```
leonidaz, line 32: weird, i'm not getting any ts errors here.

packages/ripple/src/compiler/phases/1-parse/index.js
```diff
@@ -898,6 +900,7 @@ function RipplePlugin(config) {
 					return this.parseFor(node, null);
 				}
 
+				// @ts-ignore — acorn internal: isLet accepts 0 args at runtime
```
leonidaz, line 903: we already have a type on the `isLet()` method in the parse.d.ts so we can add this to the top of the method definition, around line 762, i included the same method description:

```js
			/**
			 * Override isLet to recognize `let &{` and `let &[` as variable declarations.
			 * Acorn's isLet checks the char after `let` and only recognizes `{`, `[`, or identifiers.
			 * The `&` char (38) is not in that set, so `let &{...}` would not be parsed as a declaration.
			 * @type {Parse.Parser['isLet']}
			 */
```

packages/ripple/src/runtime/internal/client/events.js
```diff
@@ -168,7 +168,7 @@ export function handle_event_propagation(event) {
 				null;
 
 			try {
-				var delegated = current_target['__' + event_name];
+				var delegated = /** @type {Record<string, any>} */ (current_target)['__' + event_name];
```
leonidaz, line 171: i'm also not seeing any issues here as the type is declared above as `var current = /** @type {null | Element} */ (path[0] || event.target);`

and we're inside the loop where current cannot be null which TS sees.

packages/ripple/src/compiler/phases/1-parse/index.js
```diff
@@ -940,7 +943,9 @@ function RipplePlugin(config) {
 				}
 
 				let containsEsc = this.containsEsc;
-				let refDestructuringErrors = new DestructuringErrors();
+				let refDestructuringErrors = new /** @type {new () => Parse.DestructuringErrors} */ (
```
leonidaz, line 946: also, strange, i don't see any issues here with refDestructuringErrors:

`let refDestructuringErrors: DestructuringErrors`

PR #863, opened 2026-04-17
fix children prop ordering when spreads follow explicit children

RazinShafayet2007:
## Problem
Composite component transforms currently reorder explicit `children` props when a spread follows them. A call like:
```tsx
<Card children={fallback} {...props} />
```
should preserve left-to-right prop precedence, allowing the later spread to override children. Instead, the generated output effectively moves children to the end, so fallback incorrectly wins in both client and SSR output.
## Fix
- push explicit `children` props into the generated props list at their original source position
- continue merging implicit child content into that same prop with `??`
- only synthesize a trailing `children` prop when no explicit `children` prop was provided
- add client and server regression tests covering compile output and runtime rendering

---

> [!NOTE]
> **Medium Risk**
> Touches the component transform in both client and SSR compilers, which can change generated output and runtime rendering for any component usage that mixes `children` and spread props.
> 
> **Overview**
> Fixes composite component codegen so an explicit `children` attribute is emitted **in source order** relative to later `{...spreads}`, restoring normal left-to-right prop precedence (so later spreads can override `children`).
> 
> Updates both client and server transforms to only append a synthesized trailing `children` prop when no explicit `children` was provided, while still merging implicit child content into the same `children` value via `??`.
> 
> Adds regression tests for client + SSR covering both compiled output ordering and runtime rendering, and includes a patch changeset entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 628a72c1a61d65c32a5b328eec8ba524308d6394. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/tasty-mirrors-tease.md
+++ b/.changeset/tasty-mirrors-tease.md
@@ -2,4 +2,4 @@
 'ripple': patch
 ---
 
-fix: children prop ordering when spreads follow explicit children
+fix: preserve children prop precedence after spreads
--- a/.changeset/tasty-mirrors-tease.md
+++ b/.changeset/tasty-mirrors-tease.md
@@ -1,0 +1,5 @@
+---
+'ripple': patch
+---
+
+fix: children prop ordering when spreads follow explicit children
--- a/packages/ripple/src/compiler/phases/3-transform/client/index.js
+++ b/packages/ripple/src/compiler/phases/3-transform/client/index.js
@@ -1659,6 +1659,7 @@ const visitors = {
 			const props = [];
 			/** @type {AST.Property | null} */
 			let children_prop = null;
+			let has_explicit_children_prop = false;
 
 			for (const attr of node.attributes) {
 				if (attr.type === 'Attribute') {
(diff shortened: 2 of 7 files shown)


leonidaz:
I think the implicit children should override the explicit, that's what both React and Solid do.  While the spread after an explicit children should override the explicit one or vice versa - purely by order.

The order of precedence (highest to lowest)
1. Implicit JSX children (inside the tag) → ALWAYS wins
2. Last children in props (respecting spread order)
3. Earlier children props

**Case 1**
```jsx
<Component children="A" {...{ children: "B" }} />
```

➡ Result: "B"
(spread comes after → overrides)

**Case 2**
```jsx
<Component {...{ children: "B" }} children="A" />
```

➡ Result: "A"
(explicit after spread → overrides)

**Case 3**
```jsx
<Component children="A" {...{ children: "B" }}>
  C
</Component>
```

➡ Result: "C"

@trueadm what's your take?

trueadm:
Yep, we should follow what React etc do here. Also maybe hold off on this merge, as I'm about to refactor the compiler logic.

RazinShafayet2007:
closing in favor of #864

PR #864, opened 2026-04-17
Alternative to #863 - children handling fix

leonidaz:
Alternative to #863 to fix the children order precedence

---

> [!NOTE]
> **Medium Risk**
> Touches core compiler output for component calls on both client and SSR paths; subtle precedence/merge changes could affect existing apps that relied on the previous `children` fallback behavior.
> 
> **Overview**
> Fixes component invocation `children` precedence in both client and server transforms so **template children always override** any explicit `children=` prop, rather than being merged via `??`.
> 
> Also tightens compilation behavior around explicit/spread `children`: explicit `children` is always wrapped with `normalize_children`, spreads that may include `children` are routed through `spread_props` (client) / preserved as spreads (server) to maintain JSX-like last-write-wins ordering, and the `Children` type is widened accordingly. Adds focused client/server runtime and compiler tests covering explicit vs spread vs template-children ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68c8448173776ebed263e8f41becc31a1cf7a871. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trueadm:
Can we update this given the change please, then I can do a release?

leonidaz:
> Can we update this given the change please, then I can do a release?

yeah, i'll do it now

trueadm (approved):

PR #886, opened 2026-04-20
feat: serialization and hydration for trackAsync

leonidaz:
> [!NOTE]
> **High Risk**
> High risk because it changes core SSR/client hydration and reactive tracking APIs (`track`, `derived`, `trackAsync`) and introduces new error-routing/serialization paths that could affect suspense boundaries and rendering correctness.
> 
> **Overview**
> Enables **`trackAsync` SSR result hydration** by emitting per-call JSON `<script>` envelopes on the server (resolved payload plus direct dependency hashes, or a sanitized error message) and consuming/removing them during client hydration to avoid re-running the user async function.
> 
> Updates the client runtime to register tracked/derived values by compile-time hashes during hydration, hydrate `trackAsync` from the matching script tag, and wire future reactivity by replaying the serialized dependency reads.
> 
> Extends the compiler/analyzer to generate stable per-call hashes and pass them into the internal `track`/`track_async` runtime APIs on both server and client, and adds server-side `trackAsync` error routing/serialization (DEV shows real messages; prod uses a safe generic message) with script-safe escaping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5255b82d2254b45982653bd0dad6590825d5ec05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/ripple/src/runtime/internal/server/blocks.js
```diff
@@ -90,6 +98,15 @@ export function try_block(try_fn, catch_fn = null, pending_fn = null) {
 		if (created_block.f & TRY_CATCH_BLOCK) {
 			created_block.o.clear();
 			cancel_async_operations(created_block);
+
+			// make sure to serialize trackAsync error so the client can hydrate them properly
+			// needs to happen after clearing output
+			if (error instanceof TrackAsyncRunError) {
+				var { tracked: t, cause } = /** @type {InstanceType<typeof TrackAsyncRunError>} */ (error);
+				serialize_track_async_error(t.h, error);
```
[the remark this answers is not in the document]
leonidaz: @cursor the reason for this is to send down an error message that doesn't expose server internals, so it was on purpose, what do you suggest then is the best handle it?

packages/ripple/src/runtime/internal/server/index.js
```diff
@@ -1077,30 +1153,102 @@ export function exclude_from_object(obj, exclude_keys) {
...
+
+/**
+ * @param {(str: string) => void} push_fn
+ * @param {string} hash
+ * @param {object} envelope - The envelope containing the serialized data
+ * @envelope {ok: boolean, payload?: any, error?: { message: string } }
+ * @returns {void}
+ */
+function push_script_for_hydration(push_fn, hash, envelope) {
+	var serialized_envelope = JSON.stringify(envelope).replace(/</g, '\\u003C');
```
leonidaz: @cursor 

replace the the call to `.replace` with a function:

```js

function escape_script(str) {
  return str.replace(/</g, '\\u003c').replace(/>/g, '\\u003e');
}
```

which you place inside packages/ripple/src/utils/escaping.js and import it here

packages/ripple/tests/hydration/compiled/server/return.js
```diff
@@ -1957,10 +1957,10 @@ export function ReactiveDeepNestedIndependentReturns() {
 	var __r_44 = false;
 	var __r_45 = false;
 	var __r_46 = false;
-	let lazy_7 = _$_.track(false);
-	let lazy_8 = _$_.track(false);
-	let lazy_9 = _$_.track(false);
-	let lazy_10 = _$_.track(false);
+	let lazy_7 = _$_.track(false, void 0, void 0, '0222c312');
```
trueadm: I think the is should be the 2nd argument if it's going to be this common now. Otherwise we get code explosion with all the `void 0`

trueadm (approved):
Looks good other than my comment which we should fix :)

packages/ripple/tests/hydration/compiled/server/return.js
leonidaz: done, i moved them for both client and server, so that get and set are at the end

PR #923, opened 2026-04-22
fix: preserve Volar mappings for explicit call type arguments

### LINKED ISSUE #920: [Bug]: Volar crashes on function call with explicit type parameters
### Issue Category

Compilation Bug

### Bug Description

```tsx
component Test() {
  const [foo, setFoo] = useState<string | null>(null)
}
```

That syntax fails with `No source map entry for position "2:47"`.

### Checklist

RazinShafayet2007:
Fixes #920 

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only relax Volar source-map mapping generation for TypeScript generic parameter nodes to skip missing whole-node mappings instead of throwing, plus a regression test.
> 
> **Overview**
> Fixes Volar mapping generation for TypeScript generics by using `maybe_get_mapping_from_node` for `TSTypeParameterInstantiation`/`TSTypeParameterDeclaration` and skipping mappings when the source map lacks coverage, preventing crashes on calls with explicit type arguments (e.g. `useState<string | null>(...)`).
> 
> Adds a regression test ensuring `compile_to_volar_mappings` does not throw for explicit call type arguments, and includes a patch changeset for `@tsrx/core` and `@tsrx/react`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3ac9a1f8caaeff53a31bbc6492967769cce6d61. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/tsrx/src/transform/segments.js
```diff
@@ -1618,9 +1622,18 @@ export function convert_source_map_to_mappings(
...
-						get_mapping_from_node(node, src_to_gen_map, gen_line_offsets, mapping_data_verify_only),
+					// Generic spans can be emitted by downstream transforms with sparse source-map
+					// coverage around the angle-bracket delimiters. Skip missing whole-node mappings
+					// instead of crashing Volar, and rely on child type-node mappings instead.
+					const mapping = maybe_get_mapping_from_node(
+						node,
+						src_to_gen_map,
+						gen_line_offsets,
+						mapping_data_verify_only,
 					);
```
leonidaz, line 1629: is this actually true in this case?  it's not clear if this is the fix or the moving of node.callee up.

RazinShafayet2007: I think this is the correct fix. I tried to solve this with `node.callee` reordering, but that still resulted in the same error. Forgot to keep the order in the previous state. Reverting that.

trueadm (approved):
LGTM

PR #925, opened 2026-04-22
Fix ineffective attribute break detection for inline object literals

trueadm:
## Summary
- fix attribute-break detection in `printElement` so forced opening-tag breaks are triggered for multiline-prone attribute values in the reported case
- replace the ineffective `willBreak(attrDoc)`-only check with a combined approach:
  - keep `willBreak(attrDoc)` for docs that already contain unconditional breaks
  - add AST-based detection for breakable attribute values
- refine the AST heuristic so **short top-level ternary attributes stay inline when they fit**, while nested ternaries inside object-literal attribute values still trigger forced attribute breaking
- add a regression test for the reported case:
  - `<div class={selected === 0 ? 'selected' : ''}>{`div 1`}</div>` remains single-line when within `printWidth`
- remove unreachable `TSRXExpression` handling from attribute breakability logic and narrow the helper type to `AST.Expression`
- update affected prettier-plugin tests to run at `printWidth: 200` so they validate forced-break behavior instead of width-overflow side effects
- add a changeset for `@ripple-ts/prettier-plugin`
- apply formatter output updates to impacted Ripple test fixtures so `pnpm format:check` passes in CI

## Why
In the merged PR scenario, single-line source object literals in attributes are printed via grouped docs containing conditional `line` breaks, so `willBreak(attrDoc)` remains `false` and the forced-break path stays inactive.

A naive `canBreak(attrDoc)` replacement over-matches and causes unrelated formatting regressions. The final logic narrows the force-break trigger to multiline-prone object-literal attribute values while preserving inline formatting for simple, width-fitting top-level ternary attributes.

The follow-up cleanup removes an unreachable branch that suggested attribute values could be `TSRXExpression`, which is not true for `Attribute.value` in this AST.

## Validation
- reproduced the original issue at wide width (`printWidth: 200`) before fixing
- reproduced and fixed reviewer-reported regression for short top-level ternary attributes
- `pnpm format:check`
- `pnpm test --project prettier-plugin -- packages/prettier-plugin/src/index.test.js`

<div><a href="https://cursor.com/agents/bc-319c9a58-cc14-42f4-920e-41b2d69790f9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-319c9a58-cc14-42f4-920e-41b2d69790f9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


trueadm:
@cursor `<div class={selected === 0 ? 'selected' : ''}>{`div 1`}</div>` can fit onto a single line in terms of width though, why wrap it to the next line? That isn't right.

trueadm:
@cursor address feedback

PR #960, opened 2026-04-26
Fix narrowing after rendered early returns

### LINKED ISSUE #959: [Bug]: Expected type narrowing but didn't get it
### Issue Category

Compilation Bug

### Bug Description

Expected `foo` type to be narrowed to `string` after the `if` block.

Note: Type narrowing works when I remove the JSX inside the `if` block but keep the `return`.

Image

### Reproduction Steps

```tsx
declare function getFoo(): string | null;

export component Test() {
  const foo = getFoo()

  if (!foo) {
    <div>{"Foo not found"}</div>
    return
  }

  <div>{foo.trim()}</div>
}
```

### Environment

```markdown
@tsrx/preact 0.0.7
vscode-plugin 1.0.8
```

### Additional Context

_No response_

### Checklist

aleclarson:
When an early-return branch rendered JSX before returning, the shared JSX transform treated the whole if statement as render control flow and generated an inner IIFE return, so TypeScript could not narrow values after the branch. This broadens the early-return detection to branches ending in bare return and reuses the existing render-statement builder for the branch body, preserving rendered fallback output while emitting an outer component return that TypeScript can narrow across.

Closes #959 

## Before
```tsx
declare function getFoo(): string | null;

export function Test() {
	const foo = getFoo();

	return <>{(() => {
			if (!foo) {
				return <div>{"Foo not found"}</div>;
			}

			return null;
		})()}<div>{foo.trim()}</div></>;
}
```

## After
```tsx
declare function getFoo(): string | null;

const Test__static1 = <div>{"Foo not found"}</div>;

export function Test() {
	const foo = getFoo();

	if (!foo) {
		return Test__static1;
	}

	return <div>{foo.trim()}</div>;
}
```

---

> [!NOTE]
> **Medium Risk**
> Changes core JSX transform control-flow rewriting for early-return `if` statements and JSX capture sequencing, which could affect generated output across platforms if edge cases regress.
> 
> **Overview**
> Preserves TypeScript narrowing across early-return `if` branches that *render fallback JSX* by broadening early-return detection to any `if` consequent that ends with a bare `return`, and by emitting the branch’s rendered output via `build_render_statements` instead of wrapping the `if` in an inner IIFE/fragment return.
> 
> Updates JSX child capture tracking to share a single counter (`CaptureState`) across outer and nested early-return branches to avoid shadowing/redeclaring captured JSX variables, and adds Preact coverage for both narrowing and capture behavior. Includes a patch changeset for `@tsrx/core`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b4448bbde1abb85cc1a9946f2b76f9630fdc400. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/early-return-narrowing.md
+++ b/.changeset/early-return-narrowing.md
@@ -1,0 +1,5 @@
+---
+"@tsrx/core": patch
+---
+
+Preserve TypeScript narrowing after early-return branches that render fallback JSX.
--- a/packages/tsrx-preact/tests/basic.test.js
+++ b/packages/tsrx-preact/tests/basic.test.js
@@ -13,6 +13,54 @@ runSharedSourceMappingTests({
 runSharedCompileTests({ compile, name: 'preact', classAttrName: 'class' });
 
 describe('@tsrx/preact basic', () => {
+	it('preserves outer early-return narrowing when the branch renders before returning', () => {
+		const { code } = compile(
+			`declare function getFoo(): string | null;
+
+			export component Test() {
+				const foo = getFoo();
+
+				if (!foo) {
+					<div>{"Foo not found"}</div>
+					return;
+				}
+
+				<div>{foo.trim()}</div>
+			}`,
+			'Test.tsrx',
+		);
+
+		expect(code).toContain('if (!foo) {');
+		expect(code).toContain('return Test__static1;');
+		expect(code).toContain('return <div>{foo.trim()}</div>;');
+		expect(code).not.toContain('return <>{(() =>');
+	});
+
+	it('does not shadow outer JSX captures in nested rendered early-return branches', () => {
+		const { code } = compile(
+			`export component App() {
+				let value = 'outer';
+				<p>{value}</p>
+				value = 'after outer';
+
+				if (value) {
+					<span>{value}</span>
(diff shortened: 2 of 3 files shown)


leonidaz:
this wouldn't work because of hooks-safety react rule, like if you had it like this:

```jsx
import { useEffect } from 'react';

declare function getFoo(): string | null;

export component App() {
	const foo = getFoo();

	if (!foo) {
		<div>{'Foo not found'}</div>
		return;
	}

	useEffect(() => {
		console.log(foo);
	}, [foo]);

	<div>{foo.trim()}</div>
}
```

trueadm:
We should make early returns works with conditional hooks by breaking them out to separate components like we do for control flow

leonidaz:
> We should make early returns works with conditional hooks by breaking them out to separate components like we do for control flow

Yeah, tried that but we still have the problem with types as we're passing in props to the components but we need to type them at the component declaration level and we can't do it, not reliably.

But I think I finally figured out a solution...

aleclarson:
I assume @leonidaz will take it from here!

PR #978, opened 2026-04-26
fix: blocks should be allowed in function inside components

leonidaz:
closes #965 

same fix as #979 

---

> [!NOTE]
> **Medium Risk**
> Changes TSRX parsing/tokenization for component bodies to distinguish template braces from real JS blocks inside nested callables; parser regressions could affect JSX/generic disambiguation in edge cases.
> 
> **Overview**
> Fixes a parsing bug where `{ ... }` blocks inside functions/arrow functions/methods declared within a `component` body could be misinterpreted as template expression containers.
> 
> The TSRX Acorn plugin now tracks nested function/body and block depth, temporarily unwinds template (`tc_expr`) contexts while parsing JS blocks, and tightens the `<` disambiguation so generic-looking arrow functions inside nested callables still parse correctly.
> 
> Adds new client/server/compiler/runtime regression tests across Ripple and the Vite React/Solid plugins, and includes a patch changeset for `@tsrx/core`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c92a1bead7db0fea78cea111efda8b8598604b48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #983, opened 2026-04-27
Fix blocks simpler

leonidaz:
alternative to #978 

---

> [!NOTE]
> **Medium Risk**
> Touches core TSRX parsing/tokenization around `{}` blocks and `<` disambiguation inside components, so regressions could affect template parsing or TypeScript generic arrows. Added tests reduce risk but coverage is limited to the new cases.
> 
> **Overview**
> Fixes TSRX parsing so `{ ... }` inside functions/arrow functions/class methods declared within a component body is parsed as a normal JavaScript block (preserving block scope) rather than being treated as a template expression container.
> 
> To support this, the Acorn plugin now tracks nested function-body depth, gates the component-level `{}` expression-container parsing on being *outside* function bodies, and adds a conservative `<...>(...) =>` lookahead in `readToken` so generic arrow functions inside components don’t get tokenized as JSX.
> 
> Adds compile + client/server + React/Solid runtime regressions covering nested block scopes in component-local callables and ensuring component-level `{expr}` behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbfe1d868429623d9130fd3dc18073fb53c855d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.vscode/settings.json
```diff
@@ -39,5 +39,5 @@
   "[typescriptreact]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "cSpell.words": ["tsrx"]
+  "cSpell.words": ["tsrx", "tstc"]
```
trueadm, line 42: What's tstc?

leonidaz: Collection of consts in our parser. I started adding words here that we use that vscode doesn’t recognize as spelled correctly and underlines them. It’s been an annoyance.

PR #987, opened 2026-04-27
Support double quoted static string

### LINKED ISSUE #761: Allow quoted string literals without wrapping them in braces inside a template.
Twitter exchange on this issue with Dominic https://x.com/trueadm/status/2048421812564832525?s=20

```
export default component Counter() {
  <div class="container">
    <h2>{'Counter'}</h2>
    let &[count] = track(0);
    <p>{`Count: ${count}`}</p>
    let &[double] = track(() => count * 2);
    <p>{`Double: ${double}`}</p>
    <button onClick={() => count--}>{'-'}</button>
    <button onClick={() => count++}>{'+'}</button>
    if (count !== 0) {
      <div><button onClick={() => count = 0}>{'Reset'}</button></div>
    }
  </div>
}
```
There can only be two kinds of syntax within a template, 
1) A child template that starts and ends with `<`, `>` symbol.
2) Text to be displayed represented by a quoted string literal and/or any javascript expression that resolves to a string within `{` and `}` symbols.
3) A combination of 1 and 2.

Currently, To display text, the quoted string literals have to be placed inside braces within a template. This can be a bit annoying to type out the braces for all such instances of displaying text, which can be a lot in a typical component or a page. It’s also a bit noisy to see so many brace characters on the editor screen.

## Requested feature

**If there is an expression/syntax that starts and ends with quotes between the template’s `<`, `>` symbols, cant the compiler automatically assume that its a string to be displayed? AFAIK, this syntax of quoted string is not ambiguous and does not collide with other type of expressions/syntax.**

so 
`<p>"hello world"</p>`
`<p>"clicked " {count} " times"</p>`  should be possible for the compiler to parse.   

Variable references or JS expressions should be placed inside braces while quoted strings can be placed without braces. This rule already applies when assigning values to props on a template.

trueadm:
Fixes https://github.com/Ripple-TS/ripple/issues/761#issue-3997427858

---

> [!NOTE]
> **Medium Risk**
> Introduces new parsing/tokenization behavior inside `component` template bodies, which can subtly affect JSX/tag vs TypeScript-generic disambiguation and downstream compilation for existing `.tsrx` code.
> 
> **Overview**
> Adds support for writing **static text directly as double-quoted children** in TSRX template bodies (e.g. `<p>"Clicked " {count} " times"</p>`), lowering it to the existing `Text` node shape.
> 
> Updates the TSRX parser to recognize double-quoted string tokens in template-child position and to allow adjacent tag starts immediately after such text, adds compile tests for the new syntax (and a regression for compact string comparisons), and refreshes docs/examples across the repo to prefer the new quoting style. Also ships a changeset bumping `@tsrx/core` (minor) and includes an implementation plan doc.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bb0ceff196ac8b13a14cec28b30fa17d26cc23d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/double-quoted-text-children.md
+++ b/.changeset/double-quoted-text-children.md
@@ -1,0 +1,5 @@
+---
+"@tsrx/core": minor
+---
+
+Allow direct double-quoted static text children in TSRX templates.
--- a/README.md
+++ b/README.md
@@ -131,8 +131,8 @@ export component App() {
   let &[count] = track(0);
 
   <div>
-    <p>{"Count: "}{count}</p>
-    <button onClick={() => count++}>{"Increment"}</button>
+    <p>"Count: "{count}</p>
+    <button onClick={() => count++}>"Increment"</button>
   </div>
 }
 ```
@@ -160,7 +160,7 @@ export component App() {
   const count = track(0);
 
   <div>{count.value}</div>
-  <button onClick={() => count.value++}>{"Increment"}</button>
+  <button onClick={() => count.value++}>"Increment"</button>
 }
 ```
 
@@ -179,10 +179,10 @@ export component App() {
   let &[quadruple] = track(() => double * 2);
 
   <div>
-    <p>{"Count: "}{count}</p>
-    <p>{"Double: "}{double}</p>
-    <p>{"Quadruple: "}{quadruple}</p>
-    <button onClick={() => count++}>{"Increment"}</button>
+    <p>"Count: "{count}</p>
+    <p>"Double: "{double}</p>
+    <p>"Quadruple: "{quadruple}</p>
+    <button onClick={() => count++}>"Increment"</button>
   </div>
 }
 ```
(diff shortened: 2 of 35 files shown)


sreekar339339:
Instead of treating double quoted string literal as a special case of syntax and semantics, the compiler in theory can automatically wrap the single, double, backtick quoted strings inside the braces i.e pre-judge that these 3 kinds of strings are syntaxes that have to be handled as though they are in expression container braces and proceed. The differentiations between static and reactive strings is unnecessary because we treat all the 3 kinds of strings as though they belong inside expression containers just as before this feature request.

The programmer had to wrap the single, double, backtick strings in braces before. This seems like repetitive task with no special cases that the compiler can automate without creating a special case of synyax and semantics for just double quoted strings and leave out single and backtick strings.

trueadm:
> Instead of treating double quoted string literal as a special case of syntax and semantics, the compiler in theory can automatically wrap the single, double, backtick quoted strings inside the braces i.e pre-judge that these 3 kinds of strings are syntaxes that have to be handled as though they are in expression container braces and proceed. The differentiations between static and reactive strings is unnecessary because we treat all the 3 kinds of strings as though they belong inside expression containers just as before this feature request.
> 
> The programmer had to wrap the single, double, backtick strings in braces before. This seems like repetitive task with no special cases that the compiler can automate without creating a special case of synyax and semantics for just double quoted strings and leave out single and backtick strings.

I was going for symmetry with JSX attribute string values. I didn't want to overcomplicate it by supporting all the permutations, just as you can't with JSX attribute values.

trueadm:
The one thing that might be slightly confusing is this case:

```jsx
<div>
  "this is text"
</div>
```

```jsx
const get_text = () => "this is text";

<div>
  get_text()
</div>
```

Maybe this will trip up some LLMs

leonidaz:
> ```js
> const get_text = () => "this is text";
> 
> <div>
>   get_text()
> </div>
> ```
> 
> Maybe this will trip up some LLMs

i wonder how, it's clearly a function and should inside braces.
```jsx
<div>
   {get_text()}
</div>
```

PR #988, opened 2026-04-27
fix(tsrx): handle Volar source map gaps

aleclarson:
- `SwitchStatement` mapping tried to use the whole statement span, but the source map has no entry for the `switch` keyword; now Volar visits the discriminant and cases without requiring that broad mapping.
- Generic type parameter spans can be missing entries around angle brackets; now those whole-span mappings are skipped when sparse while child type mappings still run.
- Added shared source-mapping repros so React, Preact, and Solid cover both cases.
- Validated with `pnpm vitest run --project tsrx-react --project tsrx-preact --project tsrx-solid packages/tsrx-react/tests/basic.test.js packages/tsrx-preact/tests/basic.test.js packages/tsrx-solid/tests/basic.test.js`.

---

> [!NOTE]
> **Medium Risk**
> Touches the core Volar mapping walker (`segments.js`), which can affect editor diagnostics/navigation across many syntax forms. Changes are localized and guarded, but incorrect mapping behavior could regress IDE features.
> 
> **Overview**
> Fixes Volar source-mapping edge cases that could crash or mis-map positions.
> 
> `SwitchStatement` no longer emits a whole-node mapping (relying on child-node mappings instead), and generic wrapper nodes (`TSTypeParameterInstantiation`/`Declaration`) now catch and ignore missing source-map positions instead of throwing.
> 
> Adds shared no-crash repro tests for `SwitchStatement` and an arrow-function return type inside a call argument, and includes a patch changeset for `@tsrx/core`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b8bfd602fe1a42b55c22ebdbc2f0929e80ca3a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tsrx/src/transform/segments.js
+++ b/packages/tsrx/src/transform/segments.js
@@ -1699,14 +1699,21 @@ export function convert_source_map_to_mappings(
 					// Generic spans can be emitted by downstream transforms with sparse source-map
 					// coverage around the angle-bracket delimiters. Skip missing whole-node mappings
 					// instead of crashing Volar, and rely on child type-node mappings instead.
-					const mapping = get_mapping_from_node(
-						node,
-						src_to_gen_map,
-						gen_line_offsets,
-						mapping_data_verify_only,
-					);
-					if (!(mapping instanceof Error)) {
+					try {
+						const mapping = get_mapping_from_node(
+							node,
+							src_to_gen_map,
+							gen_line_offsets,
+							mapping_data_verify_only,
+						);
 						mappings.push(mapping);
+					} catch (error) {
+						if (
+							!(error instanceof Error) ||
+							!error.message.startsWith('No source map entry for position')
+						) {
+							throw error;
+						}
 					}
 				}
 				// Generic type parameters - visit to collect type variable names
--- a/packages/tsrx/src/transform/segments.js
+++ b/packages/tsrx/src/transform/segments.js
@@ -1450,12 +1450,6 @@ export function convert_source_map_to_mappings(
 					}
 				}
 
-				if (node.loc) {
-					mappings.push(
-						get_mapping_from_node(node, src_to_gen_map, gen_line_offsets, mapping_data_verify_only),
-					);
-				}
-
 				return;
 			} else if (node.type === 'SwitchCase') {
 				// Visit in source order: test, consequent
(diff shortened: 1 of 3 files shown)


trueadm (approved):

packages/tsrx/src/transform/segments.js
```diff
@@ -1705,14 +1699,21 @@ export function convert_source_map_to_mappings(
...
+						);
 						mappings.push(mapping);
+					} catch (error) {
+						if (
+							!(error instanceof Error) ||
+							!error.message.startsWith('No source map entry for position')
+						) {
+							throw error;
+						}
 					}
```
leonidaz, line 1717: this is likely not the right fix which was on top another wrong fix.  I'm going to to dig into this to figure out the missing or extra source mappings.

just fyi for the future: we need to be fixing actual source map generation in transformers or in segments if we're missing a mapping.

aleclarson: @leonidaz I was also suspicious of that, and I asked GPT5 about it and this was its explanation:
> **Me:** do your fixes accidentally swallow unsupported syntax, making it more difficult to debug Volar mapping gaps in the future?
>
> **GPT5:** Short answer: no for unsupported syntax, but the generic fix intentionally hides one narrow wrapper-span gap.
> 
> - `SwitchStatement`: does not swallow anything. It only removes a broad whole-node mapping for `switch (...) { ... }`. The walker still visits the discriminant, cases, and consequent statements, so missing mappings inside the switch still surface.
> 
> - Generic span fix: catches only `No source map entry...` from `get_mapping_from_node(node)` for `TSTypeParameterInstantiation` / `TSTypeParameterDeclaration`. That helper only resolves that node’s start/end positions; it does not visit children. After the catch, child type params still get visited, so unsupported child syntax or missing child mappings still throw.
> 
> The tradeoff: it does make missing whole-span mappings for generic angle-bracket wrappers quieter. That seems intentional and matches the existing comment, but the PR body should probably say “skip sparse wrapper mappings” rather than implying all generic mapping gaps are fixed.

leonidaz: we only needed the switch mapping for the diagnostics as some errors actually cannot be displayed or look broken, so other things like intellisense, etc don't show up on these, so that's why we had `mapping_data_verify_only` set of params on it.

but really it was the previous fix started shadowing actual problems.  anyway all good, you can see how I fixed this and other things in #993

PR #991, opened 2026-04-27
Fix duplicate TSRX refs for React and Preact

aleclarson:
- React and Preact TSRX lowered each `{ref ...}` to its own JSX `ref` prop, so multiple refs on one element were not represented correctly.
- Collapse repeated `{ref ...}` attributes into one generated callback that applies each ref to the same node.
- Add React and Preact compiler coverage for multiple refs on one element, plus a patch changeset.

---

> [!NOTE]
> **Medium Risk**
> Changes core JSX attribute lowering for React/Preact by generating composed `ref` callbacks (including cleanup handling), which could affect runtime ref behavior on any element using refs.
> 
> **Overview**
> Fixes incorrect output when a single element has multiple `{ref ...}` attributes by collapsing them into **one generated `ref` callback** that applies each ref to the same node (and runs any returned cleanup functions on unmount / ref reset).
> 
> Adds React and Preact compiler tests to assert the composed-ref output and introduces a patch changeset bumping `@tsrx/core`, `@tsrx/react`, and `@tsrx/preact`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e57463774e86ee45380bff1c9d4acf95837f7770. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fair-refs-compose.md
+++ b/.changeset/fair-refs-compose.md
@@ -1,0 +1,7 @@
+---
+"@tsrx/core": patch
+"@tsrx/react": patch
+"@tsrx/preact": patch
+---
+
+Compose multiple `{ref ...}` attributes into a single JSX ref callback for React and Preact targets.
--- a/packages/tsrx-preact/tests/basic.test.js
+++ b/packages/tsrx-preact/tests/basic.test.js
@@ -82,6 +82,29 @@ describe('@tsrx/preact basic', () => {
 		expect(code).toContain("{'preact tsx'}");
 	});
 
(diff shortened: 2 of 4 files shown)


aleclarson:
Addressed this in the latest push:

- Function ref return values are captured.
- The composed ref returns a cleanup callback when any function ref returns one.
- Mixed refs still tear down: no-cleanup callbacks receive `null`, and object refs are reset to `null` from the composite cleanup.

PR #994, opened 2026-04-27
fix: tsrx - handling multiple refs with multiple styles

leonidaz:
alternative to #991 

- merge tsrx-style with platform-style refs
- for React and Preact use a small runtime `mergeRefs` helper
- fix the issue in Solid for merging solid-style and tsrx-style refs

---

> [!NOTE]
> **Medium Risk**
> Changes core JSX attribute lowering and ref semantics across multiple targets (React/Preact/Solid/Vue) and adjusts Ripple runtime ref dispatch, which could affect generated output and mount/unmount behavior in user apps.
> 
> **Overview**
> **Adds compile-time multi-ref merging across TSRX targets and tightens `ref` validation.** The core transform now rejects elements with multiple TSX-style `ref={...}` attributes, but allows any number of keyword-form `{ref expr}` refs and merges them with at most one `ref={...}`.
> 
> React/Preact/Vue now collapse multiple refs into `ref={__mergeRefs(...)}` with an injected `mergeRefs` helper (new `@tsrx/<target>/merge-refs` export backed by `@tsrx/core/runtime/merge-refs`), while Solid collapses to `ref={[...]}`. Vue specifically switches from an inline callback merge to the shared helper to support ref objects and mixed `ref={...}` + `{ref ...}` cases.
> 
> Ripple’s runtime `ref()` wiring is expanded to dispatch between function refs, `Tracked`-style ref objects, and plain assignment targets, with added tests and docs covering the updated ref matrix and behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1843fb665d96c8ef3abf237c90c2a64f8c7ce1f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/tsrx-solid/src/transform.js
```diff
@@ -1309,39 +1322,7 @@ function transform_element_attributes(raw_attrs, is_composite, transform_context
...
-		/** @type {any} */ ({
-			type: 'JSXAttribute',
-			name: { type: 'JSXIdentifier', name: 'ref', metadata: { path: [] } },
-			value: to_jsx_expression_container(argument),
-			shorthand: false,
-			metadata: { path: [] },
-		}),
-		source_node,
-	);
+	return mergeDuplicateRefs(result, /** @type {any} */ (transform_context));
```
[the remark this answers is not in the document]
leonidaz: a false positive:

transform_element_attributes_dispatch (where the second merge_duplicate_refs call lives) is only reached from core's to_jsx_element and create_dynamic_jsx_element — both bypassed by Solid.

transformElement instead calls Solid's local to_jsx_element, which calls Solid's local transform_element_attributes. That local helper is the only path where Solid attributes get merged.

packages/tsrx-react/tests/basic.test.js
```diff
@@ -1676,4 +1676,51 @@ describe('lazy destructuring', () => {
...
+			expect(code).not.toContain('__mergeRefs');
+			expect(code).not.toContain('@tsrx/react/merge-refs');
+		});
+
+		it('collapses multiple ref={expr} attributes into a __mergeRefs call and injects an import', () => {
+			const { code } = compile(
+				`export component App() {
+					function refA(_node) {}
+					function refB(_node) {}
+					<div ref={refA} ref={refB}>{'hi'}</div>
```
aleclarson: shouldn't this be `{ref refA} {ref refB}`?

leonidaz: I thought we wanted to support multiple refs no matter if it's an attribute or a tsrx ref expression.  As they were compiled to the same output and we were allowing multiple ref attributes - thought we're making ref a special case basically.  But if we're going to allow only one ref attribute `ref={func}` and multiple ref expressions {ref a}, I'm fine with that.

@trueadm your call.

I guess it's also related to how https://github.com/Ripple-TS/ripple/issues/996 should be handled.

trueadm: We only want to support one ref prop, but allow multiple {ref expr}. We should also support variable refs like we do Ripple

leonidaz: we don't support variables for refs in ripple or anywhere, we only support functions.

And the ref attribute, `ref={whatever}` - this would be the native platform's ref (except for Ripple that doesn't have it), right?  we don't do anything with it so you can use with `useRef()` in react for example?

trueadm: I meant mutable refs as per https://github.com/Ripple-TS/ripple/issues/996

trueadm (approved):

leonidaz:
i'm adding handling for ripple, almost done - what we discussed: handling tracked and mutable vars

PR #995, opened 2026-04-28
docs(features): document render props

### LINKED ISSUE #926: [Docs]: Guidance around “Render props”
I see no mention of the render-props pattern in the docs. It'd be great to have guidance on how to handle that when migrating to TSRX!

Thanks

aleclarson:
- Add target-neutral guidance in the `/features` Children section explaining render props as function-valued props.
- Include a small `render={...}` example showing JSX wrapped in a fragment.
- Closes #926


website-tsrx/src/pages/features.tsrx
```diff
@@ -626,6 +636,21 @@ export component FeaturesPage() {
 						<pre class="code-block">
 							<code>{html CHILDREN_HTML}</code>
 						</pre>
+						<p class="section-body">
+							<strong>{'React/Preact note: '}</strong>
```
leonidaz: shouldn't this be a universal case, not tied in to any platform?

trueadm (approved):

PR #1002, opened 2026-04-28
Update double quote string heursitcs

trueadm:
> [!NOTE]
> **Medium Risk**
> Updates the TSRX tokenizer/parser for double-quoted text children, which can change how templates compile/render (entities, newlines, and quote termination). Formatter and grammar updates are low risk, but parser changes may affect edge-case parsing and backward compatibility.
> 
> **Overview**
> Direct double-quoted TSRX text children are now tokenized like quoted JSX attribute values: **character references are decoded**, newlines are handled, and **backslashes no longer act as escapes** (unterminated quotes now error instead of accepting JS-style escapes).
> 
> The parser now preserves the authored quoted form via a `TextNode.raw` field, and the Prettier plugin prints that raw text directly and adjusts single-child inlining heuristics accordingly.
> 
> Updates the TextMate grammar to highlight direct quoted text as JSX text, and adds/extends compile + client/server/runtime tests and docs/spec to lock in the new entity/backslash behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb8bac9651ab5b0852cb56c0509cdc3cfa76d562. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trueadm:
@sreekar339339 Does this affect your recent changes any?

trueadm:
@leonidaz I'm not sure about this PR but wanted to put it up for discussion. It changes the double quote JSX text in TSRX to be more like JSXText, in that you have to use HTML entities rather than string escaping rules. Let me know your thoughts.

leonidaz:
I like this change.  It makes it easier to write html (especially multiline) and more in line with user expectations.  If people still need the js escaping they can use {text }

So definitely more pros than cons, let's ship it.

leonidaz:
the only thing I noticed is the grammar syntax support, not sure if it's doable.

like the word `else` inside double quotes would be highlighted like the js else.

and prettier needs fixing as it convert it to {text }

trueadm:
> the only thing I noticed is the grammar syntax support, not sure if it's doable.
> 
> like the word `else` inside double quotes would be highlighted like the js else.
> 
> and prettier needs fixing as it convert it to {text }

Good points, let me fix them as part of this PR.

PR #1025, opened 2026-05-01
fix: stale outer binding when control flow block apply changes

leonidaz:
The code after control flows should have the values of the shared top variables set inside the flows.  This only focused on fixing React and Preact.  Solid and Vue, I think should not have these issues but I only checked if statements for Solid.  Ripple definitely cannot have these issues.

I haven't checked the volar editor support to make sure things are mapped to the correct generated items.  I'll do that in my morning.

```tsx
export component App({ show }: { show: boolean }) {
	let x: number | undefined;
	if (!show) {
		[x] = useState(100);

		<div>{x}</div>
	}
	console.log(x); // should print 100 and not undefined
}
```

```tsx
export component App({ kind }: { kind: 'a' | 'b' }) {
	let x: number | undefined;
	kind = 'b';
	switch (kind) {
		case 'a':
			[x] = useState(100);
			<div>{x}</div>
			break;
		case 'b':
			[x] = useState(200);
			<span>{x}</span>
			break;
	}
	console.log(x); // should print 200 and not undefined
```

```tsx
// obviously not something one should do lol
export component App({ items }: { items: string[] }) {
	let last: number | undefined;

	for (const item of [1, 2, 3]; index index; key item) {
		[last] = useState(item);

		<div>{last}</div>
	}
	console.log(last); / should print 3 and not undefined
}
```

```tsx
let cache: number | undefined;
const dataPromise = new Promise<number>((resolve) => setTimeout(() => resolve(42), 2000)).then(
	(v) => {
		cache = v;
	},
);

const loadData = () => {
	if (cache !== undefined) return cache;
	throw dataPromise;
};

export component Test({ load }: { load: () => number }) {
	let data: number | undefined;
	try {
		[data] = useState(load());
		<div>{data}</div>
	} pending {
		<p>{'loading'}</p> // should show 'loading' first for a couple of seconds
	} catch (err) {
		<div>{'error'}</div>
	}
	console.log(data); // should show 42 vs undefined
}

export component App() {
	<Test load={loadData} />
}
```

---

> [!NOTE]
> **Medium Risk**
> Changes core JSX transform control-flow rewriting for React/Preact (if/switch/try/for-of) and alters generated output shapes; mistakes could break render semantics (fall-through, error boundaries, iteration) or source mapping.
> 
> **Overview**
> Fixes a React/Preact-only bug where hooks inside control-flow blocks (`if`, `switch`, `try`/`pending`/`catch`, `for-of`) could mutate outer bindings but code *after* the block would still see stale values.
> 
> This introduces a **continuation-lift** transform in `packages/tsrx/src/transform/jsx/index.js`: it generates a “tail helper” component for the statements following a hook-bearing control-flow block and ensures each branch/case/iteration invokes that tail so post-hook values flow into subsequent code (with special handling for `switch` fall-through/break-only cases and empty-iteration `for-of`).
> 
> `for-of` hook bodies are additionally hoisted: the loop helper and TS type aliases move outside the loop, the iteration source is normalized via `Array.isArray(...) ? ... : Array.from(...)`, and tail execution is gated on the last iteration.
> 
> Adds extensive new snapshot coverage for these scenarios (React/Preact), updates a React test assertion, and refactors various manually-built JSX AST nodes to use new builder helpers (`jsx_opening_element`, `jsx_closing_element`, `jsx_element_fresh`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d00699bea60ce815152cd1f75ff506475b0fc7e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/smooth-icons-shave.md
+++ b/.changeset/smooth-icons-shave.md
@@ -1,0 +1,18 @@
+---
+'@tsrx/preact': patch
+'@tsrx/react': patch
+'@tsrx/react-playground': patch
+'@tsrx/solid-playground': patch
+'@tsrx/core': patch
+---
+
+Fixes a bug where for all control statements: for, if, switch, try/pending/catch
+where using hooks inside to change values, like useState, would not be reflected
+in the subsequent code. The fix involved creating continuation hooks and calling
+them at the end of the control flow block - it's an oversimplification.
+
+Fixes the for loop by hoisting the generated statement body hooks and types to the
+outside of the loop.
+
+Refactors a bunch, but not all, manually created AST nodes into using ast builder
+functions.
--- a/packages/tsrx-preact/tests/__snapshots__/basic.test.js.snap
+++ b/packages/tsrx-preact/tests/__snapshots__/basic.test.js.snap
@@ -224,17 +224,19 @@ export function App({ kind }: { kind: 'a' | 'b' }) {
 		return null;
 	});
 
+	const _tsrx_StatementBodyHook2_x = x;
+
+	const StatementBodyHook2 = App__StatementBodyHook2 ?? (App__StatementBodyHook2 = function StatementBodyHook2({ x }: { x: typeof _tsrx_StatementBodyHook2_x }) {
+		[x] = useState(100);
+
+		return <><div>{x}</div><StatementBodyHook1 x={x} /></>;
+	});
+
 	switch (kind) {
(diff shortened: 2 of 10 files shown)


trueadm (approved):
Awesome

PR #1032, opened 2026-05-01
fix: always hoist loop body with hooks + coerce to array

leonidaz:
> [!NOTE]
> **Medium Risk**
> Changes core JSX transform behavior for `for...of` by introducing new hoisting/scoping mechanics and always wrapping iteration sources with `Array.isArray`/`Array.from`, which could affect generated code order and runtime semantics across React/Preact outputs.
> 
> **Overview**
> Fixes `for...of` codegen so **hook-bearing loop bodies are always hoisted** (including when the loop appears as a JSX child), avoiding per-iteration helper rebinding and ensuring nested loop setup stays in the correct scope.
> 
> Adds a `function_scope_statements` side-channel to safely inject hoisted helper/normalization statements just before the component/helper `return` (preserving user-declaration order and preventing TDZ issues), and **normalizes all default `for...of` iteration sources** to arrays via `Array.isArray(x) ? x : Array.from(x)`.
> 
> Extends hoisted-loop typing to support destructured value params by deriving leaf prop types via TS indexed-access chains, with a fallback path for unsupported patterns (e.g. `...rest`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7a2f39d1147c1b5128e70acd6351fbb03ea4b92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/all-lemons-sit.md
+++ b/.changeset/all-lemons-sit.md
@@ -1,0 +1,8 @@
+---
+'@tsrx/preact': patch
+'@tsrx/react': patch
+'@tsrx/core': patch
+---
+
+Always hoist generated body above loops with hooks.
+Always coerce iterable expression to array via Array.from if not an array
--- a/.changeset/bright-masks-review.md
+++ b/.changeset/bright-masks-review.md
@@ -1,0 +1,5 @@
+---
+"@tsrx/mcp": minor
+---
+
+Add TSRX authoring review tools for accessibility, scoped style usage, and component structure.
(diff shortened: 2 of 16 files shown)


trueadm (approved):

leonidaz:
this was a draft, i need to redo it, separating types compile from prod.  And you already fixed the hosting for the loop.

leonidaz:
changed direction since to stop allowing hooks modifying var outside the block

PR #1042, opened 2026-05-02
Multiple bug fixes

trueadm:
> [!NOTE]
> **Medium Risk**
> Changes core compilation/validation and target transforms for `for...of` rendering, which can affect generated code and runtime behavior across multiple frameworks. Risk is mitigated by extensive new shared/unit/runtime tests covering `continue`, keying, and loop validation rules.
> 
> **Overview**
> **Component `for...of` control flow is now consistent across TSRX targets.** `continue` is supported to skip a rendered iteration, while top-level `return`/`break` inside component `for...of` loops and non-`for...of` loop statements in template scope now produce targeted diagnostics.
> 
> **Transforms were updated to preserve correct output semantics.** `continue` is lowered safely (e.g., callback `return`/skip placeholders rather than emitting JS `continue`), loop key handling was tightened (including keyed fragments requiring `Fragment` imports for React/Preact), and Vue’s loop lowering uses `flatMap` when the body can render `null` to truly skip items.
> 
> **Tooling/docs/test coverage expanded.** MCP adds new advice for loop diagnostics and docs are updated to describe the rules; VS Code grammar highlights TSRX inside anonymous `component` expressions; parser fixes text-only fragment initializers before `{expr}` children; new shared regression tests and runtime tests (including a new jsdom suite for `vite-plugin-preact`) validate the behavior end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c73bc36f5a8cfd832047f6df6bb5cff19a74c3ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/tsrx/tests/shared/compile.js
```diff
@@ -55,6 +55,101 @@ export function runSharedCompileDiagnosticsTests({ compile_to_volar_mappings, na
...
 }
 
+/**
+ * Shared component-loop regressions. Vue does not share the full JSX output
+ * suite because its component export shape differs, but it should still share
+ * these component-body validation rules.
+ *
+ * @param {Pick<CompileHarness, 'compile' | 'name'>} harness
+ */
+export function runSharedComponentLoopControlFlowTests({ compile, name }) {
```
leonidaz, line 65: we need a test for a loop with a hook, e.g. useState, that has some markup after the loop that reads the variable that's declared above the loop and the loop sets inside.  The part after the loop gets put into its own function and it runs on the last loop iteration at the end.  This continuation was done for account for the hook behavior and that scoping bug.

So, with the `continue` statement, if it's the last loop iteration, we need to call that "after the loop" function before we exit with the return.  Maybe it already works that way, i haven't checked.  But bringing this up.

leonidaz: like so:

```jsx
export component App() {
  let last: number | undefined;
  for (const item of [1, 2, 3]; index i) {
    if (i === 2) continue;
    [last] = useState(item);
    <div key={i}>{last}</div>
  }
  console.log(last);
}
```

packages/tsrx/src/transform/jsx/index.js
```diff
@@ -3180,12 +3531,107 @@ function apply_key_to_render_statements(statements, key_expression) {
...
+ * @returns {any}
+ */
+function create_jsx_key_attribute(key_expression) {
+	return /** @type {any} */ ({
+		type: 'JSXAttribute',
+		name: { type: 'JSXIdentifier', name: 'key', metadata: { path: [] } },
+		value: to_jsx_expression_container(clone_expression_node(key_expression), key_expression),
+		metadata: { path: [] },
+	});
+}
```
leonidaz: already have this as a function in builders

packages/tsrx/src/transform/jsx/index.js
```diff
@@ -3180,12 +3531,107 @@ function apply_key_to_render_statements(statements, key_expression) {
...
+		},
+		closingElement: {
+			type: 'JSXClosingElement',
+			name: clone_jsx_name(name),
+			metadata: { path: [] },
+		},
+		children: fragment.children,
+		metadata: { path: [] },
+	});
+}
```
leonidaz, line 3618: we have this in builders.js
jsx_element_fresh
jsx_closing_element
jsx_opening_element

PR #1054, opened 2026-05-03
hoist React branch components to module level scope

trueadm:
> [!NOTE]
> **Medium Risk**
> Changes React code generation for hook-bearing control-flow branches/loop bodies by emitting module-scope helper components instead of cached per-render helpers, which can affect runtime output shape and downstream tooling expectations. Editor (Volar) output is kept on the old typed-cached path via a new option, reducing but not eliminating risk of regressions in edge cases.
> 
> **Overview**
> React hook-isolation helpers extracted from control flow (and hook-bearing element child statement bodies) are now emitted as **module-scope component declarations** (e.g. `function App__StatementBodyHookN(...)`) rather than lazily created/cached helpers inside the parent component, and the generated call sites are updated accordingly.
> 
> The shared JSX transform adds a `moduleScopedHookComponents` switch (wired through `JsxTransformContext` and platform hooks) that skips emitting synthetic helper prop type aliases/`typeof`-based typing when module-scoping is enabled, avoiding the prior synthetic `any` prop annotations.
> 
> `@tsrx/react` enables module-scoped hook components for runtime compilation, while `compile_to_volar_mappings` explicitly disables it so Volar output preserves the prior lexical typed-cached helper pattern for editor tooling; tests/snapshots are updated and shared regex expectations are loosened to accept the new helper naming.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75af7d7201bc61c21b0dbe9ddb53047e0e1c7b88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/tsrx-react/tests/basic.test.js
```diff
@@ -378,18 +379,18 @@ describe('@tsrx/react basic', () => {
 				mapping.lengths[0] === 'foo'.length,
 		);
 
-		expect(code).toContain('let App__StatementBodyHook1;');
-		expect(code).toContain('let App__StatementBodyHook2;');
+		expect(code).toContain('function App__StatementBodyHook1()');
+		expect(code).toContain('function App__StatementBodyHook2({ foo }: { foo: any })');
```
leonidaz: so generating of type any definitely seems not right, as the type should be from getFoo() - where it should be discovered and used via typeof, I think we create an intermediate var for the type but it wouldn't work via doing it with foo since we're hoisting the statement body hook into the module, so we'd have to then base it on getFoo since it came outside the component. But if getFoo were to be inside the component then we'd have to hoist getFoo to the module level otherwise we won't be able to get the type at the module level (types also follow lexical scopes and we can't read a type inside)

Anyway, this gets kind of hairy but probably doable.  Otherwise, we'd need to create a separate transformation to support types, like we do for Ripple.  Might be easier as we can just compile as is for types.

trueadm: Applied a better fix

leonidaz (approved):
I think this is good to merge, we can fine tune the types later and restructure for types if needed.

PR #1061, opened 2026-05-03
more fixes

trueadm:
> [!NOTE]
> **Medium Risk**
> Touches core and Solid JSX lowering logic to treat certain standalone expressions as render output, which could subtly change codegen for non-JSX expression statements in templates across targets.
> 
> **Overview**
> Fixes a codegen bug where fragment shorthand with a single expression child (e.g. `<> {value} </>`) could be lowered as an executable statement instead of render output when it appears directly in component bodies or inside `if`/`for...of`/`switch` render bodies.
> 
> This adds a new shared predicate `is_bare_render_expression` and uses it in both the core JSX transform and the Solid transform to wrap these expressions in `JSXExpressionContainer` so they participate in the returned render fragment, plus adds shared regression tests and a changeset bumping all framework packages as patch releases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e1b37c5fea3ee3e91fe2c93ba39d905b629ccde. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tsrx-solid/src/transform.js
+++ b/packages/tsrx-solid/src/transform.js
@@ -290,6 +290,8 @@ function component_to_function_declaration(component, transform_context) {
 			} else {
 				render_nodes.push(jsx);
 			}
+		} else if (is_bare_render_expression(child)) {
+			render_nodes.push(to_jsx_expression_container(child, child));
 		} else {
 			statements.push(child);
 		}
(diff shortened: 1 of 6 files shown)


packages/tsrx-solid/src/transform.js
```diff
@@ -765,6 +771,53 @@ function is_null_literal(node) {
...
+		case 'TSSatisfiesExpression':
+		case 'TSNonNullExpression':
+		case 'UnaryExpression':
+		case 'UpdateExpression':
+		case 'YieldExpression':
+			return true;
+		default:
+			return false;
+	}
+}
```
leonidaz: is this identical to the one in packages/tsrx/src/transform/jsx/index.js? utils?

PR #1095, opened 2026-05-08
fix: more <tsrx> parser expression fixes

### LINKED ISSUE #1094: Parser rejects <tsrx> returned from render callback props
## Summary

A `<tsrx>...</tsrx>` expression compiles when returned from a standalone arrow-function block, but the same callback shape fails when used as a JSX render callback prop.

## Reproduction test case

```js
it('lowers native TSRX template fragments returned from render callback props', () => {
	const { code } = compile(
		`class Foo { bar() { return <List render={() => { return <tsrx><span>"Item"</span></tsrx>; }} />; } }`,
		'App.tsrx',
	);

	expect(code).toContain('{"Item"}');
	expect(code).not.toContain('<tsrx>');
});
```

## Actual result

Compiling that source fails with:

```text
JSX elements cannot be used as expressions. Wrap JSX with `<>...</>` or `<tsx>...</tsx>`, wrap TSRX templates with `<tsrx>...</tsrx>`, or use elements as statements within a component.
```

## Expected result

The callback return expression should be accepted and the `<tsrx>` wrapper should be lowered, matching standalone callback behavior:

```ts
const render = () => { return <tsrx><span>"Item"</span></tsrx>; };
```

That standalone form currently compiles to a returned `<span>{"Item"}</span>`.

## Notes

The same JSX-prop parser context also rejects related callback forms, including `() => <tsrx>...</tsrx>` and `() => { return <tsx>...</tsx>; }`, so the issue appears to be tied to parsing expression islands inside JSX attribute expressions.

leonidaz:
fixes #1094 
fixes other identified use cases

---

> [!NOTE]
> **Medium Risk**
> Touches low-level parser token-context stack handling for JSX/TSRX boundaries; mistakes here can cause broad parse regressions across many syntactic forms.
> 
> **Overview**
> Fixes `<tsrx>` parsing when used as an *expression* by more aggressively cleaning up Acorn token contexts after parsing a template element, including closing `}`, `)`, and `]` cases and a callback-return shape.
> 
> Adds regression tests covering `<tsrx>` used in JSX attribute expressions when parenthesized, passed as a call argument, nested in an object literal, and returned from a render-prop callback, and publishes the change as a patch via a changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 056f7b1fc7f3a8c30f0bc20dd268ee3a0953f6e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/vast-areas-allow.md
+++ b/.changeset/vast-areas-allow.md
@@ -1,0 +1,8 @@
+---
+'@tsrx/core': patch
+---
+
+Parser fix for <tsrx> - cleans up the pending token context for }, ), ], plus the callback-return case:
+parenthesized: content={(<tsrx>...</tsrx>)}
+passed as a call arg: content={wrap(<tsrx>...</tsrx>)}
+used as an object property: content={{ child: <tsrx>...</tsrx> }}
--- a/packages/tsrx/src/plugin.js
+++ b/packages/tsrx/src/plugin.js
@@ -320,15 +320,20 @@ export function TSRXPlugin(config) {
 				}
 			}
 
-			#popJsxAttributeExpressionContextAfterTemplateElement() {
-				if (this.type !== tt.braceR) {
+			#popTokenContextsAfterTemplateExpressionElement() {
+				const context_index = this.context.length - 1;
+				if (
+					this.context[context_index] === b_stat &&
+					this.context[context_index - 1] === tstc.tc_expr
+				) {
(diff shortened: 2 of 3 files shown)


aleclarson:
🎉 🎉 🎉 🎉

PR #1125, opened 2026-05-12
Remove tsrx-tsc compile fallback

aleclarson:
Removes the extra `compile()` fallback after `compile_to_volar_mappings()` fails.

When a TSRX module fails to parse, using fallback generated code can move the failure downstream into modules that import it. It is clearer to let the broken TSRX module take the diagnostics directly instead of risking follow-on errors in consumers.

---

> [!NOTE]
> **Medium Risk**
> Changes failure-mode behavior for `tsrx-tsc` by no longer generating placeholder TS output, which may alter diagnostics/emitted errors for broken `.tsrx` modules and their dependents.
> 
> **Overview**
> Stops `tsrx-tsc` from attempting a secondary runtime `compile()` when `compile_to_volar_mappings()` throws, and removes the associated `getFallbackGeneratedCode` helper.
> 
> On compile failure, the plugin now consistently keeps the raw `.tsrx` source as the generated virtual code (with 1:1 mappings/verification) and only logs the fatal error in `TSRX_TSC` mode, rather than producing best-effort TS-valid fallback output.
> 
> Adds a patch changeset documenting the behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e32d2f0ba0efc039982478af807326c07e74eba3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leonidaz:
this needs to be here because if we fail to compile, we want to point the error to the exact source location instead of pointing it at the top of the file or some some random place.  

If you have a particular use case, please create an issue and share the minimal reproduction so we can see if we can improve this.

aleclarson:
@leonidaz No that's not right. The branch I removed is only used by `tsrx-tsc`. Just today, I tried using `tsrx-tsc` and the `compile()` fallback was ineffective.

leonidaz:
ah, sorry my bad, didn't see that it's for the tsrx-tsc

ok, if it doesn't work. let's remove it.

leonidaz (approved):

PR #1127, opened 2026-05-12
fix(eslint-parser): use tsrx core directly

aleclarson:
Switches @tsrx/eslint-parser from @tsrx/ripple to @tsrx/core and removes the ripple peer dependency. Also removes the unused loader module and updates the parser README to describe TSRX rather than Ripple. Open question: should this parser still support any Ripple-specific syntax beyond TSRX? If yes, we should decide how that should be wired, possibly via an optional peer dependency or a separate parser entrypoint.

---

> [!NOTE]
> **Medium Risk**
> Swaps the underlying parser implementation and dependency surface (removing `ripple` peer dep), which can change AST shape/locations and thus lint rule behavior for `.tsrx` files.
> 
> **Overview**
> `@tsrx/eslint-parser` now parses `.tsrx` files by calling `@tsrx/core`’s `parseModule` directly instead of dynamically loading `@tsrx/ripple`, and removes the global/require-based Ripple compiler loader code.
> 
> Package metadata/build config is updated accordingly: `ripple` is dropped from peer deps, `@tsrx/core` is added as a dependency and marked external for bundling, and docs are rewritten to describe TSRX usage (including updated install instructions and file globs).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 847b461a4fabcc45a9bc99485515970a4b38c064. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leonidaz:
Ripple's syntax would be included in @tsrx/core and if we end up adding specific eslint rules for Ripple for some Ripple-only syntax, then we can figure out the wiring.  But the goal is to keep syntax on par as much as possible.

We also want to move to oxlint

After we get this merged, a similar cleanup would be really nice for the eslint-plugin

packages/eslint-parser/README.md
```diff
@@ -108,21 +104,14 @@ The parser will successfully parse this and allow ESLint rules (like those from
...
-
 ## Related Packages
 
 - [@tsrx/eslint-plugin](https://www.npmjs.com/package/@tsrx/eslint-plugin) -
-  ESLint rules for Ripple
-- [ripple](https://ripplejs.com) - The Ripple framework
+  ESLint rules for TSRX
 - [@ripple-ts/vite-plugin](https://www.npmjs.com/package/@ripple-ts/vite-plugin) -
-  Vite plugin for Ripple
+  Vite plugin for TSRX
```
leonidaz: i think we can remove this, it's not really related and we have a plugin for each platform target

leonidaz (approved):
👍

PR #1160, opened 2026-05-20
fix(prettier-plugin): print TypeScript satisfies expressions

aleclarson:
Fixes the Prettier plugin’s handling of `TSSatisfiesExpression` nodes. Previously, expressions like `props.status satisfies never` fell through to the unknown-node fallback and formatted as `/* Unknown: TSSatisfiesExpression */`.

Changes:
- Adds printer support for `TSSatisfiesExpression`.
- Adds a regression test for a switch exhaustiveness default case.


leonidaz (approved):
👍

PR #1177, opened 2026-05-28
Remove component keyword, cleanup TSRX as an expression, `{html expr}`, `{text expr}`, `{ref expr}` and `{style class}`

trueadm:
> [!NOTE]
> **High Risk**
> Breaking language and authoring-model changes across parsers, compilers, and Ripple rendering semantics affect all `.tsrx` consumers and multi-target TSRX packages.
> 
> **Overview**
> This PR **migrates Ripple/TSRX authoring to native markup in ordinary functions** and drops several legacy syntax forms. The `component` keyword is removed; UI is defined with `export function App() { return <>...</>; }` and inline control flow inside returned TSRX, as reflected in **README**, the **LiveCodes playground**, and **changesets** that also describe cross-target parser/compile/format/lint updates.
> 
> **Grammars and editor support** are aligned with that model: TextMate (`ripple.tmLanguage` / JSON) and **tree-sitter** no longer parse `component` declarations, special `<tsrx>` islands, or brace directives `{html}`, `{text}`, `{ref}`, and `{style ...}`; bare `<>...</>` fragments are first-class, and JSX children can include full statements where the grammar allows.
> 
> **Documented runtime/compiler behavior** (via changesets, not all shown in the diff) includes treating tags/fragments as default TSRX, compiling functions that return TSRX through runtime component helpers, preserving direct PascalCase helper calls, adding **`Fragment`** and **`innerHTML`** (including from spreads), and using normal `{expr}` / target-specific raw HTML props instead of removed template directives.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d7d8b6ba4c13e8b48f172b17426c5da947f25266. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sreekar339339:
why?

trueadm:
> why?

Agents are really struggling with the semantics of not having a return statement for JSX-like content and it's been causing very odd issues where both humans and machines are getting confused over how things are being displayed and how things work when debugging.

PR #1188, opened 2026-06-01
explore adding @Component decorator

trueadm:
This makes it easier for libraries like Ripple, that need explicit boundaries for static analysis to work better without the magic of an uppercase function name instead.

---

> [!NOTE]
> **High Risk**
> Changes core Ripple/TSRX compiler rules for what counts as a component; apps relying on old heuristics (internal TSRX, renderable non-TSRX returns without `@Component`) may compile differently.
> 
> **Overview**
> Introduces **`@Component`** from `ripple` as an explicit compiler marker for Ripple functions that should get component control-flow semantics without returning native TSRX directly (e.g. `null`, strings).
> 
> **Component detection is tightened:** Ripple treats a function as a component only when it **returns TSRX directly**, carries **`@Component`** (including aliased or namespace imports from `ripple`), or matches existing native-TSRX-return rules—not when it merely **creates** TSRX inside the body or returns other “renderable” values without the decorator. Related heuristics (template-in-body scanning, broad renderable-return promotion, nested prop callbacks auto-wrapped as components) are removed or narrowed; client/server transforms strip `@Component` and decorator-only `Component` imports from emitted JS while preserving type-only imports when needed.
> 
> **Parser/AST:** TSRX gains a generic **`Decorator`** node on functions via an Acorn plugin; ESTree types and a noop runtime **`Component`** export are added. Tests and docs/README/llms copy explain that **PascalCase is convention only**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f42f920bf5fa32fed412824d2cf406493e059e26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trueadm:
@cursor push 3d8269982f

PR #1196, opened 2026-06-02
fix(prettier-plugin): start restoring deleted prettier tests

leonidaz:
Start the process of restoring deleted tests for prettier.  Will be in draft until the new tsrx syntax if finalized and all deleted tests that still apply are added back.


leonidaz:
this was done as part of the new tsrx design.

PR #1321, opened 2026-07-06
Fix TSRX parser crash on parenthesized JSX branch with nested element children

WebEferen:
## What

Fixes a TSRX parser crash (`Unexpected token`) on a conditional/logical JSX branch that Prettier wraps in parentheses across multiple lines, when the branch contains a **nested non-self-closing child element on its own line** — i.e. the extremely common shape Prettier emits:

```tsx
{cond ? (
  <Outer>
    <Inner>…</Inner>
  </Outer>
) : null}
```

Single-line branches, self-closing/text children, `&&` branches, and depth-1 branches all parsed fine — only this specific (and Prettier-default) combination threw.

## Root cause

`#popTokenContextsAfterTemplateExpressionElement` restores the tokenizer context after a JSX element parsed in expression position. When the element has a nested non-self-closing child laid out on its own line, a JSX tokenizer context (`tc_expr`) is left **above** the enclosing `(`. The paren-pop only fires when `top?.token === '('`, so it never strips the leaked context. Left in place, the tokenizer stays in JSX-text mode after the `)`, and the ternary `:` is mis-tokenized as JSX text → the parse derails (reported back at the `?`).

An `&&` branch survives only by luck — its next token after `)` is `}`, which `tc_expr` closes correctly. A depth-1 branch survives because a non-JSX context (`${`) shields the top. Only *leaked-JSX-on-top* + *`:`-next* breaks.

## Fix

When the following token is a closing `)`/`]` and a run of leaked JSX contexts (`tc_expr`/`tc_oTag`/`tc_cTag`) sits directly above the matching `(`/`[`, strip them down to the delimiter and pop it — restoring JS-expression mode. 26 lines, guarded so it fires only in exactly this leaked-context case.

## Testing

- New regression tests in `parser.test.js` (ternary + `&&`/array-literal variants). The ternary test **fails without the fix, passes with it**.
- Full `@tsrx/core` suite stays green (**314 tests**).
- AST verified correct: `ConditionalExpression` → `JSXElement` consequent → nested child, `Literal null` alternate.
- `prettier --check` clean; `patch` changeset included.

## Motivation

Found while building a deterministic (non-AI) React→Octane source bridger — Octane consumes `@tsrx/core`'s `parseModule` via its compiler, and a bridged React component using this Prettier-standard formatting failed to compile. Companion Octane PR that consumes this fix: **octanejs/octane#23** (its `react-compat` bridger hit this on the Prettier-formatted branch of one example).

### DIFF
--- a/.changeset/fix-tsrx-conditional-jsx-parse.md
+++ b/.changeset/fix-tsrx-conditional-jsx-parse.md
@@ -1,0 +1,5 @@
+---
+'@tsrx/core': patch
+---
+
+Fix a TSRX parser crash on a conditional/logical JSX branch that Prettier wraps in parentheses across multiple lines when it contains a nested non-self-closing child element on its own line (e.g. `{cond ? (\n  <Outer>\n    <Inner/>\n  </Outer>\n) : null}`). A JSX tokenizer context leaked by the nested child sat above the enclosing `(`, so after the closing `)` the tokenizer stayed in JSX-text mode and mis-tokenized the following `:`.
--- a/packages/tsrx/src/plugin.js
+++ b/packages/tsrx/src/plugin.js
@@ -2358,6 +2358,32 @@ export function TSRXPlugin(config) {
 					ctx.length = ci - 1;
 					return;
 				}
+				// A JSX tokenizer context (`tc_expr`/`tc_oTag`/`tc_cTag`) leaked by a nested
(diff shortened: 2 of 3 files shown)


leonidaz:
thanks! had to go with a more robust / correct solution.  closing.

PR #1362, opened 2026-07-20
fix(prettier-plugin): treat rendered JSX text as whitespace-significant

trueadm:
## Problem

TSRX compilers preserve a rendered JSX text child's parsed value verbatim — Octane keeps it byte-for-byte, the Ripple target trims only the outer edges of newline-bearing runs, and only whitespace-only-with-newline runs render nothing. The printer, however, re-wrapped element bodies purely by print width:

- collapsed `<Item>\n\tKangaroo\n</Item>` ↔ `<Item>Kangaroo</Item>` (react-aria collections commit that string verbatim as the item `textValue`)
- split inline text like `<div>Hello <b>world</b>!</div>` onto separate lines
- re-indented multi-line text, collapsed repeated spaces, wrapped long text at printWidth
- dropped kept whitespace-only runs (`<a/> <b/>`, `<> <div /> </>`)

Every one of these changed the rendered textContent of the program (observed downstream in the octane repo as changed committed `textValue` strings in @octanejs/aria test fixtures).

## Fix

Rendered JSX text is now whitespace-significant, mirroring how Prettier treats sensitive HTML (`<pre>`):

- Text children print **byte-for-byte** via literal lines — never re-indented, trimmed, wrapped, or merged.
- Children are grouped into **glued segments**: any source-adjacent boundary touching rendered text is frozen.
- Layout only changes at seams the parser proves are free: boundaries with no text between children (whitespace-only-with-newline runs never reach the AST), and **elastic** boundaries where the tokenizer already skipped a newline run (after `</…>` closing tags, and after tsx-expression opening tags) — detected robustly via source-position gaps, where a break parses identically.
- Expression-only bodies still re-wrap as before (`<span>\n\t{expr}\n</span>` → `<span>{expr}</span>` is compiler-invisible).
- Comments authored inside a text run now print from the run's raw source slice, so reprinting them no longer inserts lines into the text.
- Overflow is absorbed by breaking the opening tag's attributes; text glued to its tags stays put whatever the print width.

## Verification

- **369/369 plugin tests pass.** Five old expectations encoded the semantics-changing behavior and were updated. A new `JSX text semantics preservation` suite parses each corpus case before/after formatting with `@tsrx/core`, asserts kept text-node values are byte-identical, and asserts idempotence (Kangaroo both forms, expression holes, mixed text+element, kept spaces, elastic seams, printWidth overflow, tsx-expression elements, control flow).
- Package typecheck, repo-wide `format:check`, and the tsrx-mcp consumer suite (68 tests) pass.
- **Blast-radius run over the octane repo** (prettier 3.8.4 + octane's own `@tsrx/core` 0.1.44, read-only): of 805 non-ignored `.tsrx` files, the fixed plugin rewraps exactly **1** file — a text-free cosmetic change (`<div><></></div>` breaking onto lines) — with **zero text-affecting changes**, zero errors, and full idempotence. Direct old-vs-new comparison shows 0.3.104 corrupts rendered text in four distinct shapes that this build preserves.

## Release

Patch changeset included → ships as `@tsrx/prettier-plugin@0.3.105` for the octane repo to consume.

---

> [!NOTE]
> **Medium Risk**
> Wide-reaching formatter behavior change for all `.tsrx` JSX, but scoped to printing semantics with extensive regression tests; wrong elastic/frozen boundaries could still alter rendered text.
> 
> **Overview**
> **Rendered JSX text is now printed byte-for-byte** so Prettier no longer changes compiled `textContent`. The old path trimmed, merged, re-indented, and wrapped text to `printWidth` (e.g. collapsing multi-line labels or splitting inline `"Hello <b>world</b>!"` across lines).
> 
> `JSXText` uses verbatim output via `printVerbatimJSXText`; only **droppable** whitespace-only runs with newlines print empty. Element and fragment bodies are built from **glued segments** in `collectJSXChildSegments`: boundaries that touch rendered text stay frozen, while **elastic** seams (source gaps after closing tags / certain openings, or expression-only bodies) can still break. Long lines absorb overflow by breaking attributes, not text. Comments inside a text run reprint from the raw source slice so they are not emitted on separate lines.
> 
> Tests add a **`JSX text semantics preservation`** suite (parse with `@tsrx/core`, compare text-node values, idempotence) and update expectations that previously encoded semantics-changing layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a238d463e4556d13fc469efdb930efa612e229d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/whitespace-significant-jsx-text.md
+++ b/.changeset/whitespace-significant-jsx-text.md
@@ -1,0 +1,17 @@
+---
+'@tsrx/prettier-plugin': patch
+---
+
+Treat rendered JSX text as whitespace-significant so formatting never changes
+the program. TSRX compilers keep a rendered text child's parsed value verbatim
+(only whitespace-only runs containing a newline are dropped), but the printer
+used to re-wrap element bodies purely by print width — collapsing
+`<Item>\n\tKangaroo\n</Item>` to `<Item>Kangaroo</Item>`, splitting inline
+text onto new lines, re-indenting multi-line text, and collapsing repeated
+spaces — all of which changed the rendered textContent. Rendered text is now
+reproduced byte-for-byte: boundaries that touch text are frozen, and layout
+only changes at seams the tokenizer already treats as elastic (dropped
+whitespace-only runs and skipped runs after closing tags), where re-wrapping
+cannot affect the parsed text. Comments authored inside a text run are now
+printed from the run's raw source slice so reprinting them no longer inserts
+extra lines into the text.
--- a/packages/prettier-plugin/src/index.test.js
+++ b/packages/prettier-plugin/src/index.test.js
@@ -4,6 +4,7 @@ import standalonePrettier from 'prettier/standalone';
 import estreePlugin from 'prettier/plugins/estree';
 import { fileURLToPath } from 'url';
 import { dirname, join } from 'path';
+import { parseModule } from '@tsrx/core';
 import { languages, parsers, printers } from './index.js';
 
 const __filename = fileURLToPath(import.meta.url);
@@ -506,16 +507,19 @@ const items=[1,2,3];
   </div>;
 }`;
 
+		// Rendered text is whitespace-significant: the authored line breaks and
+		// blank lines inside each text child are part of its rendered value and
+		// must survive formatting byte-for-byte.
 		const expected = `function Test() {
   return <div>
-    <p class="status">
-      Visible:
+    <p class="status">Visible:
+
       {String(visible)}
     </p>
     <p>
       {name}
-      is visible
-    </p>
+
+      is visible</p>
     <p>Hello {name}!</p>
   </div>;
 }`;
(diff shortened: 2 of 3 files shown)


packages/prettier-plugin/src/index.test.js
```diff
@@ -506,16 +507,19 @@ const items=[1,2,3];
...
     </p>
     <p>
       {name}
-      is visible
-    </p>
+
+      is visible</p>
     <p>Hello {name}!</p>
   </div>;
 }`;
```
leonidaz, line 525: this doesn't seem to be correct, actual Prettier produces the following (excluding parens that we don't do):

```tsx
function Test() {
	return <div>
		<p class="status">
			Visible:
			{String(visible)}
		</p>
		<p>
			{name}
			is visible
		</p>
		<p>Hello {name}!</p>
	</div>;
}
```

packages/prettier-plugin/src/index.test.js
```diff
@@ -5671,17 +5688,20 @@ const foo = <><Bar {...props} /></>;`;
...
       class={styles.notificationMessage}
   >The report is ready. Review the summary before sharing it with the team.</span>
 }`;
 
 			const expected = `function App() {
-  return <span class={styles.notificationMessage}>
-    The report is ready. Review the summary before sharing it with the team.
-  </span>;
+  return <span
+    class={styles.notificationMessage}
```
leonidaz, line 5703: prettier also doesn't do this

.changeset/whitespace-significant-jsx-text.md
```diff
@@ -0,0 +1,17 @@
+---
+'@tsrx/prettier-plugin': patch
+---
+
+Treat rendered JSX text as whitespace-significant so formatting never changes
+the program. TSRX compilers keep a rendered text child's parsed value verbatim
+(only whitespace-only runs containing a newline are dropped), but the printer
+used to re-wrap element bodies purely by print width — collapsing
+`<Item>\n\tKangaroo\n</Item>` to `<Item>Kangaroo</Item>`, splitting inline
```
leonidaz, line 9: this is standard prettier behavior

leonidaz (commented):
I'm not sure about these changes, they go against what the standard Prettier does.

PR #1392, opened 2026-07-27
fix(vite-plugin-react): make .tsrx imports visible to Vite's dep scanner

jamiedavenport:
Vite's dependency scanner doesn't run the main plugin pipeline, so it can't read `.tsrx` files. Any npm dependency imported only from `.tsrx` modules is invisible at dev-server startup and gets discovered at request time instead, triggering a re-optimize and a full page reload. I hit this in an app where Playwright runs specs in parallel against one dev server: a request in flight during the reload loads mixed pre/post-optimize chunks and fails with `Invalid hook call` (jamiedavenport/onyx#63). Dev-only; `vite build` runs the full pipeline and is unaffected.

## Fix

Same approach `@tsrx/vite-plugin-vue` already ships: a `config` hook registers a dep-scan plugin under `optimizeDeps.rolldownOptions.plugins` that compiles `.tsrx` to tsx for the scan pass. Two details:

- `optimizeDeps.extensions` needs `'.tsrx'`, since the scanner's catch-all resolver externalizes unknown extensions before any plugin can load them.
- The scan output explicitly imports `<jsxImportSource>/jsx-runtime`, which the main transform always emits, so the scanner records it regardless of how the scan's own jsx transform is configured.

## Tests

Unit tests for the config hook and scan transform, plus an integration test that runs a real `optimizeDeps` pass over a fixture and asserts a dep imported only from a `.tsrx` file lands in the optimized metadata. The integration test fails without the fix (the scanner finds zero deps).

---

> [!NOTE]
> **Low Risk**
> Dev-only `optimizeDeps` behavior in the Vite plugin; production builds still use the full transform pipeline, with broad test coverage for the new scan path.
> 
> **Overview**
> Fixes **dev-server startup** missing npm deps that are imported only from `.tsrx` files, which previously surfaced as a late **re-optimize + full reload** (and could break parallel tests with mixed chunk versions).
> 
> The React Vite plugin now exposes a **`config` hook** that adds `optimizeDeps.extensions: ['.tsrx']` and a Rolldown **dep-scan plugin** (`create_dep_scan_plugin`) under `optimizeDeps.rolldownOptions.plugins`. That pass **compiles `.tsrx` to TSX** for crawling, **prepends** `<jsxImportSource>/jsx-runtime`, **omits** CSS virtual imports, and on compile failure returns an **empty module** so one bad file does not abort pre-bundling for the whole app.
> 
> Adds **unit + `optimizeDeps` integration tests** (including a malformed `.tsrx` sibling) and updates **TypeScript plugin types** for the new `config` surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 975ebbc67aee549eda209946a935d69b13c8ea80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/vite-plugin-react-dep-scan.md
+++ b/.changeset/vite-plugin-react-dep-scan.md
@@ -10,4 +10,6 @@ and got discovered at request time instead, forcing a re-optimize and a full
 page reload. The plugin now registers a dep-scan plugin under
 `optimizeDeps.rolldownOptions.plugins` (plus the `.tsrx` entry in
 `optimizeDeps.extensions`) that compiles `.tsrx` modules for the scan pass, so
-their imports are crawled up front.
+their imports are crawled up front. A `.tsrx` file that fails to compile is
+skipped by the scan rather than failing it, so one malformed file no longer
+costs the project its dependency pre-bundling.
--- a/.changeset/vite-plugin-react-dep-scan.md
+++ b/.changeset/vite-plugin-react-dep-scan.md
@@ -1,0 +1,13 @@
+---
+'@tsrx/vite-plugin-react': patch
+---
+
+fix: make `.tsrx` imports visible to Vite's dependency scanner
+
+Vite's dep scanner runs through Rolldown without the main plugin pipeline, so
+any npm dependency imported only from `.tsrx` files was invisible at startup
+and got discovered at request time instead, forcing a re-optimize and a full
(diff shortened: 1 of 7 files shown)


leonidaz:
@cursor review

leonidaz (approved):
Thank you!

PR #1413, opened 2026-08-01
fix(vite-plugin-react): scan dependencies in server environments

jamiedavenport:
I made a mistake in [#1392](https://github.com/Ripple-TS/ripple/pull/1392): the dependency scanner was registered through `config()`, so it only reached the client environment.

This moves the scanner setup to `configEnvironment()` and applies it to the client plus server environments that enable dependency discovery. It also adds a named SSR environment regression test covering dependencies imported only from `.tsrx` files.

Stet is still carrying [a local workaround](https://github.com/jamiedavenport/stet/blob/main/apps/web/vite.config.ts) for its TanStack Start SSR environment.

Tested with the `vite-plugin-react` and `tsrx-react-runtime` Vitest projects, plus the affected type checks.

---

> [!NOTE]
> **Medium Risk**
> Touches Vite dev dependency discovery for client and SSR; behavior change for multi-environment apps but scoped to environments that opt into discovery, with regression tests.
> 
> **Overview**
> Fixes a regression where the `.tsrx` dependency scanner only applied to the **client** environment because it was wired through Vite's global `config()` hook.
> 
> The plugin now uses **`configEnvironment()`** and returns the same `optimizeDeps` / dep-scan rolldown setup for the client and for any environment with **`optimizeDeps.noDiscovery === false`** (typical SSR setups with discovery on). Server environments with discovery off get no scanner config. The scan options are factored into **`create_dep_scan_config`**; public types switch from `config` to **`configEnvironment`**.
> 
> The shared test harness adds **`scanEnvironmentFixture`** (dev server + per-environment initial scan), and **`scan.test.js`** covers SSR registration rules and an end-to-end check that dependencies reachable only from `.tsrx` files are discovered in a named SSR environment.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84fad06be0e72283d5f23972716f7b0a5a75f13d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/warm-pandas-scan.md
+++ b/.changeset/warm-pandas-scan.md
@@ -1,0 +1,5 @@
+---
+'@tsrx/vite-plugin-react': patch
+---
+
+Register the TSRX dependency scanner for server environments that enable dependency discovery.
--- a/packages/tsrx/tests/shared/dep-scan.js
+++ b/packages/tsrx/tests/shared/dep-scan.js
@@ -1,7 +1,7 @@
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
-import { optimizeDeps, resolveConfig } from 'vite';
+import { createServer, optimizeDeps, resolveConfig } from 'vite';
 
 const DEFAULT_INDEX_HTML = `<!doctype html>
 <html>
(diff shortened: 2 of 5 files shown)


leonidaz:
@cursor review

leonidaz (approved):
👍

PR #1423, opened 2026-08-13
fix(tsrx): stop lookahead leaking a tokenizer context

trueadm:
## The bug

A callback prop whose parameter has a **no-argument function type** fails to parse inside a template:

```tsx
export function App() @{
	<Boundary fallback={(reset: () => void) => 'retry'}>
		<Child />
	</Boundary>
}
```

```
Unexpected token `>`. Did you mean `&gt;` or `{">"}`?
```

This is an ordinary, common shape — a typed reset/close/done callback — and it parses fine everywhere else: in setup, in a child expression container, and in plain `.tsx`. Only JSX **attribute-value** position breaks.

## Root cause

Deciding whether `(` opens a function type scans ahead, and that scan is only meant to be a query.

acorn-typescript's `getCurLookaheadState()` snapshots `context` **by reference** — unlike its sibling `cloneCurLookaheadState()`, which slices it — and `setLookaheadState()` restores that same array. So any tokenizer context the scan pushes survives the "restore".

`tsIsUnambiguouslyStartOfFunctionType` is the case that bites: for an **empty** parameter list it consumes `(` (pushing a context) and returns the moment it sees `)`, never popping. One context is then left on the stack for the rest of the parse. Inside a template that shifts every later token one frame out of phase, so the `>` closing the element's opening tag is tokenized as JSX text.

A non-empty list (`(n: number) => void`) scans far enough to balance — which is exactly why only the empty-parens spellings broke:

| spelling | before |
| --- | --- |
| `(reset: () => void) => …` | ❌ |
| `(r: () => string) => …` | ❌ |
| `(r: () => Promise<void>) => …` | ❌ |
| `(r: { go: () => void }) => …` | ❌ |
| `(r: (n: number) => void) => …` | ✅ |
| `(r: (() => void)) => …` | ✅ |
| `(r: R) => …` (alias) | ✅ |

## The fix

Override `getCurLookaheadState()` to snapshot the context stack **by value**, so lookahead stays a pure query — which is what every caller already assumes. This also covers `tryParse`'s failState restore.

## Testing

- Three regression tests in `packages/tsrx/tests/utils/parser.test.js`, covering the reported shape, all four previously-leaking spellings, and the following-attribute/sibling positions a narrower fix could still get wrong. **Red-tested**: all three fail on `main` with the original error, pass with the fix.
- `pnpm test` full suite: identical to the `main` baseline — 220 passed / 4901 tests, with the same 7 pre-existing suite-collect failures in `eslint-plugin` and `vite-plugin-react` before and after.
- `pnpm typecheck` and `pnpm format:check` clean.

Found while type-annotating a downstream `.tsrx` corpus, where this was the single remaining parse failure.

---

> [!NOTE]
> **Low Risk**
> Dependency patch bump and targeted parser regression tests; no application logic or auth/data paths change.
> 
> **Overview**
> Fixes **parse failures** on JSX callback props whose parameter is typed as a **no-argument function** (e.g. `fallback={(reset: () => void) => …}`), which previously surfaced as `Unexpected token '>'` because speculative “is this a function type?” lookahead left the tokenizer context stack out of sync after empty `()`.
> 
> **@tsrx/core** pulls **`@sveltejs/acorn-typescript` `^1.0.13`** (from `1.0.11`) so lookahead snapshots restore parser state correctly; the workspace catalog, lockfile, and a **`minimumReleaseAgeExclude`** entry for that release are updated. **`esrap`** is bumped to **`^2.3.2`** across the monorepo lockfile/catalog.
> 
> Adds **`getCurLookaheadState()`** to the parser type surface in `parse.d.ts`, plus **regression tests** in `parser.test.js` for the reported `Boundary` shape, other empty-parens type spellings, and attributes/siblings after the offending prop.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d627d9c7e4cbb4668e904b2e2e186437baae9d7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/nervous-hoops-repeat.md
+++ b/.changeset/nervous-hoops-repeat.md
@@ -1,0 +1,13 @@
+---
+'@tsrx/core': patch
+---
+
+Fix a parse error on a callback prop whose parameter has a no-argument function
+type, such as `<Boundary fallback={(reset: () => void) => …}>`. Deciding whether
+`(` opens a function type scans ahead, and the scan's state snapshot aliased the
+tokenizer's context stack instead of copying it. For an empty parameter list the
+scan consumed `(` and returned at `)` without popping the context that `(`
+pushed, leaving every later token one frame out of phase — the `>` closing the
+element's opening tag was then tokenized as JSX text and reported as
+``Unexpected token `>`. Did you mean `&gt;`?``. Lookahead now snapshots the
+context stack by value, so it cannot mutate the caller's state.
--- a/.changeset/nervous-hoops-repeat.md
+++ b/.changeset/nervous-hoops-repeat.md
@@ -3,11 +3,6 @@
 ---
 
 Fix a parse error on a callback prop whose parameter has a no-argument function
-type, such as `<Boundary fallback={(reset: () => void) => …}>`. Deciding whether
-`(` opens a function type scans ahead, and the scan's state snapshot aliased the
-tokenizer's context stack instead of copying it. For an empty parameter list the
-scan consumed `(` and returned at `)` without popping the context that `(`
-pushed, leaving every later token one frame out of phase — the `>` closing the
-element's opening tag was then tokenized as JSX text and reported as
-``Unexpected token `>`. Did you mean `&gt;`?``. Lookahead now snapshots the
-context stack by value, so it cannot mutate the caller's state.
+type, such as `<Boundary fallback={(reset: () => void) => …}>`, by upgrading
+`@sveltejs/acorn-typescript` to a version that restores parser state after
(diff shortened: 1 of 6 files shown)


leonidaz (changes requested):
looks like https://github.com/sveltejs/acorn-typescript/pull/84 fixed this in the new 1.0.13

I can just upgrade it to that version if everything is good, we can merge with the acorn-typescript new version.

leonidaz (approved):

PR #1432, opened 2026-08-18
fix: correct insertion anchor in keyed for-loop multi-item insert path

chenzylab:
## Bug

`reconcile_by_key` and `reconcile_by_ref`'s pure-insertion fast path (the `j > a_end` branch, hit when the old array is fully consumed by the head/tail trim and only new items remain to insert) computes the insertion anchor like this:

```js
if (j > a_end) {
  if (j <= b_end) {
    while (j <= b_end) {
      b_val = b[j];
      var target = block_start(a_blocks, j, a_length, anchor); // bug
      b_blocks[j] = create_item(target, b_val, j, render_fn, is_indexed, true);
      j++;
    }
  }
}
```

`block_start(blocks, index, length, fallback)` looks up `blocks[index]`, so `blocks` and `index` need to live in the same index space. Here it's called with the **old** `a_blocks` array but the **new** array's loop index `j` — the two don't correspond to the same items.

When inserting a whole batch of new keyed items into the middle of a list, every item in the batch should anchor to the same DOM position: the first old block that survived the suffix-match trim (`a_blocks[a_end + 1]`). Instead, the anchor is recomputed every iteration using the growing `j`, so only the first inserted item happens to land correctly (where `a_blocks[j]` coincidentally equals `a_blocks[a_end + 1]` on the first iteration) — every subsequent item in the batch is misplaced by one slot.

### Repro

```jsx
// old: [{id:'a'}, {id:'c'}, {id:'d'}]
// new: [{id:'a'}, {id:'b1'}, {id:'b2'}, {id:'c'}, {id:'d'}]
@for (const item of items; key item.id) {
  <div class="item">{item.text}</div>
}
```

Expected DOM order after the update: `A, B1, B2, C, D`.
Actual: `A, B1, C, B2, D` — `B2` lands after `C` instead of before it.

I found this while building a Tree component that expands two new leaf nodes under a branch that isn't the last item in the flattened list — the exact "insert ≥2 keyed items, not at the end" shape that triggers the bug.

## Fix

Hoist the anchor computation out of the loop, using the old array's fixed boundary (`a_end + 1`) instead of the loop-varying new index `j`:

```js
if (j > a_end) {
  if (j <= b_end) {
    var insert_target = block_start(a_blocks, a_end + 1, a_length, anchor);
    while (j <= b_end) {
      b_val = b[j];
      b_blocks[j] = create_item(insert_target, b_val, j, render_fn, is_indexed, true);
      j++;
    }
  }
}
```

Same fix applied to the mirrored non-keyed path in `reconcile_by_ref`.

Note the reverse-insertion branches elsewhere in the same file (around line 603/610/623) already use the correct pattern — `block_start(b_blocks, next_pos, b_length, anchor)`, i.e. the **new** blocks array with a **new** index — which is why this looks like an isolated oversight in the pure-insertion fast path rather than a broader design issue.

## Testing

- Added a regression test to `packages/ripple/tests/client/for.test.tsrx` ("keyed for inserts multiple new items in the middle in correct order") that fails before this fix and passes after.
- Ran the full `packages/ripple` test suite locally: 132 test files, 1511 tests, all passing (12 pre-existing skips, unrelated).

---

> [!NOTE]
> **Low Risk**
> Narrow fix in list reconciliation fast paths with targeted tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes **wrong DOM order** when keyed or ref-based `@for` reconcilers insert **two or more new items in the middle** of a list via the pure-insert fast path (`j > a_end`).
> 
> That path used to call `block_start(a_blocks, j, …)` on every loop iteration, mixing the **old** block array with the **new** list index `j`. The anchor drifted into the matched suffix, so later inserts landed one slot too late (e.g. `A, B1, C, B2, D` instead of `A, B1, B2, C, D`).
> 
> **`reconcile_by_key`** and **`reconcile_by_ref`** now resolve **`insert_target` once** with `block_start(a_blocks, a_end + 1, …)` and reuse it for the whole insert batch. Regression tests cover keyed and ref-based middle inserts; changeset documents the patch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90f4435f41e4d3e1ed37df9b8590e99800009ef0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leonidaz (approved):
Thank you!

PR #1434, opened 2026-08-19
fix: ref-count handle_root_events per target for shared-target Portals

chenzylab:
## Bug

`handle_root_events(target)` is called once per `Portal` on mount and returns a cleanup function. Each caller was treated independently with no notion of sharing the same `target` — the cleanup unconditionally removed the delegated event listeners from `target` and reset the shared `root_target`.

When two Portals both mount to the same target — very common, since `document.body` is the typical Portal target for things like modals, drawers/side sheets, tooltips, etc. — closing/unmounting the first Portal runs its cleanup, which tears down the delegated listeners for `document.body` entirely. Any sibling Portal still open on that same target silently loses all event handling: clicks (and other delegated events) inside it stop working with no error, no warning. The DOM nodes and their delegated handler references (`__click` etc., set in `create_event`) are completely untouched — only the shared root delegation listener on `target` is gone — which makes this very hard to track down from either component's own code.

### Repro

```jsx
function App() @{
  let &[modalOpen] = track(true);
  let &[sheetOpen] = track(true);
  <>
    @if (modalOpen) {
      <Portal target={document.body}>
        <button onClick={() => (modalOpen = false)}>{'Close modal'}</button>
      </Portal>
    }
    @if (sheetOpen) {
      <Portal target={document.body}>
        <button onClick={() => console.log('clicked')}>{'Sheet button'}</button>
      </Portal>
    }
  </>
}
```

Click "Close modal", then click "Sheet button" — nothing happens. The sheet's button still exists in the DOM with its click handler attached via the delegated-events mechanism, but `document.body` no longer has a `click` listener routing to it, because the modal Portal's cleanup removed it when it unmounted.

I ran into this building a Modal + SideSheet pair for a component library, both portaling to `document.body` — closing the Modal silently broke click handling in an already-open SideSheet.

## Fix

`handle_root_events` now ref-counts callers per `target`:

```js
/** @type {Map<Element, { count: number, registered_events: Set<string> }>} */
var root_target_refs = new Map();

export function handle_root_events(target) {
  var ref = root_target_refs.get(target);
  if (ref === undefined) {
    ref = { count: 0, registered_events: new Set() };
    root_target_refs.set(target, ref);
  }
  ref.count += 1;
  // ... registered_events now comes from the shared ref, not a fresh Set per call

  return () => {
    // ...
    var current_ref = root_target_refs.get(target);
    current_ref.count -= 1;
    if (current_ref.count > 0) return; // other callers still need this target's listeners
    // last caller: actually tear down
  };
}
```

The delegated listeners for a given target are only actually torn down once every caller for that target has released it. `registered_events` is now shared per-target too (previously each call had its own `Set`, so a second Portal on the same target that introduced a new event type wouldn't correctly track it for removal purposes either).

## Testing

- Added a regression test to `packages/ripple/tests/client/portal.test.tsrx` ("two portals sharing the same target: closing one does not break click delegation for the other") — fails before this fix (`expected '0' to be '1'`, i.e. the click never reaches the handler) and passes after.
- Ran `packages/ripple` test suite locally: all passing (pre-existing failures elsewhere in the monorepo are `adapter-vercel` Node-version-support checks and `eslint-plugin` config loading, both unrelated and reproducible on `main` without this change).

---

> [!NOTE]
> **Medium Risk**
> Touches core client event delegation and Portal mount/cleanup paths; behavior change is intentional but affects all delegated events at root targets.
> 
> **Overview**
> Fixes **silent loss of delegated clicks** when multiple `Portal`s (or mounts) share the same delegation target—especially `document.body`—because one unmount used to remove the shared root listeners for everyone.
> 
> **`handle_root_events`** now keeps a per-target ref count and shared `registered_events` set; cleanup only removes listeners when the last caller for that element releases. The global **`root_target`** is replaced by **`root_target_refs`**, and **`on()`** uses **`is_root_target_or_above`** so listeners on any active root target stay non-delegated even when another Portal roots elsewhere.
> 
> **`Portal`** acquires root delegation in a **separate `render` block** tied to `props.target`, so children-only updates no longer tear down and re-register delegation on every re-render.
> 
> Regression coverage: shared-target portals, portal retargeting, and `on(container)` with a sibling Portal layer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b5bb7dcc99f74ecaa4630707358d4d9a32e341d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/chubby-lemons-clean.md
+++ b/.changeset/chubby-lemons-clean.md
@@ -2,8 +2,8 @@
 'ripple': patch
 ---
 
-Fix event delegation breaking for a Portal when a sibling Portal sharing the same target unmounts.
+Fix event delegation breaking when multiple roots (Portals, mounts) share or mix targets.
 
-`handle_root_events(target)` had no notion of multiple callers sharing the same `target` element. Each Portal (or the app root) calls it once on mount and returns a cleanup function; that cleanup unconditionally removed the delegated event listeners from `target` and reset the shared `root_target` state. When two Portals both mount to `document.body` (a very common case — e.g. a Modal and a SideSheet both portaling there), closing/unmounting the first one would tear down the delegated `click` (and other) listeners for `document.body` entirely, silently breaking every click inside the second Portal even though it was still open. The DOM nodes and their `__click` handler references were untouched — only the root delegation listener was gone — so the failure had no error, no warning, and no obvious cause from either component's own code.
-
-`handle_root_events` now ref-counts callers per `target` (`root_target_refs: Map<Element, { count, registered_events }>`). The delegated listeners for a given target are only actually torn down once every caller for that target has released it.
+- `handle_root_events(target)` had no notion of multiple callers sharing the same `target` element. Each Portal (or the app root) calls it once on mount; its cleanup unconditionally removed the delegated event listeners from `target`. When two Portals both mount to `document.body` (e.g. a Modal and a SideSheet), closing the first tore down the delegated listeners for `document.body` entirely, silently breaking every click inside the second — no error, no warning. `handle_root_events` now ref-counts callers per target, and the delegated listeners are only torn down once every caller for that target has released it.
+- The single `root_target` global is gone. `on()` now checks the element against every active root target, so attaching a listener directly to one root's target while another root (e.g. a Portal to a sibling layer) was acquired later no longer silently takes the broken delegated path.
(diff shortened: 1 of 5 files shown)


leonidaz (approved):
thank you!

PR #1459, opened 2026-09-09
fix(dom): add implicit px suffix to bare numeric style values

chenzylab:
## Summary

- `apply_styles` (in `packages/ripple/src/runtime/internal/client/render.js`) serializes every style object value with a plain `String(value)` before calling `style.setProperty(prop, value)`.
- For CSS properties that require a unit (`width`, `height`, `top`, `left`, `margin`, ...), the browser silently drops a `setProperty` call whose value is a bare unitless number — no error, no console warning, the property just never lands on the element.
- `{ width: 400 }` is an ergonomic, very common shape — and the exact one the `style` JSX attribute's `string | number` type invites — so this silently no-ops one of the most common ways to set a dynamic dimension from reactive state. Only properties that are legitimately unitless (`z-index`, `opacity`, `line-height`, ...) happen to work today, which makes the failure mode easy to miss until someone tries a plain numeric `width`/`height`.
- This mirrors what every mainstream framework already does: React, Vue, and Preact each ship a small "is this property unitless" allowlist and append `px` to everything else for numeric style values.

## Fix

Adds a `UNITLESS_NUMBER_PROPERTIES` allowlist (trimmed to properties reachable through plain `style.setProperty`, based on React's `isUnitlessNumber`) and a `format_style_value(css_prop, raw_value)` helper. `apply_styles` now routes every style value through it instead of a bare `String(...)` call. `0` is left unitless regardless of property (a bare `0` is valid CSS for length properties, and avoids an unnecessary `0px` vs `0` diff on every re-render).

## Test plan

- [x] Added `render bare numbers in a style object with an implicit px suffix` to `packages/ripple/tests/client/basic/basic.attributes.test.tsrx`, covering a `width` number getting `px`, a reactive update, and `z-index`/`opacity` staying unitless.
- [x] `npx vitest run --project ripple-client basic.attributes` — 29/29 passing.
- [x] `npx vitest run --project ripple-client` (full client suite) — 89 files, 1034 passing, 4 skipped, 0 failing.
- [x] `npx prettier --check` on both touched files.

Found while building a component library on top of Ripple: a `SideSheet`-style component passing a numeric `width` prop straight through to an inline `style` object rendered with no width at all, with no error anywhere in the pipeline.

---

> [!NOTE]
> **Low Risk**
> Targeted runtime formatting for inline style objects with broad test coverage; behavior change only affects numeric values not on the unitless allowlist.
> 
> **Overview**
> **Numeric style object values** (e.g. `{ width: 400 }`) were serialized with plain `String(...)`, which browsers ignore for length-like properties—so dimensions silently failed on both client updates and SSR markup.
> 
> This PR routes style values through a shared **`format_style_value`** helper (React-style unitless allowlist, vendor-prefixed variants, `0` stays unitless, `--*` custom properties never get `px`). **Client** `apply_styles` and **server** `get_styles` both use it so hydration stays aligned. Client and server attribute tests cover `px` suffixing, unitless props (`z-index`, `opacity`, `scale`, `-webkit-line-clamp`), and custom properties.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82d03de07080266ababaa3c2dfab7e7acd249456. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/style-numeric-unit-suffix.md
+++ b/.changeset/style-numeric-unit-suffix.md
@@ -1,0 +1,5 @@
+---
+'ripple': patch
+---
+
+Fix numeric values in a style object (e.g. `{ width: 400 }`) being silently dropped for CSS properties that require a unit. `apply_styles` serialized every value with a plain `String(...)` before calling `style.setProperty`, and the browser ignores a `setProperty` call whose value is a bare unitless number for properties like `width`, `height`, `top`, `left`, or `margin` — no error, the property just never lands on the element. Numeric style values are now formatted through an `isUnitlessNumber`-style allowlist (aligned with React's), appending `px` to any property not on it; `0` stays unitless regardless of property.
--- a/packages/ripple/src/runtime/internal/client/render.js
+++ b/packages/ripple/src/runtime/internal/client/render.js
@@ -99,6 +99,83 @@ export function set_attribute(element, attribute, value) {
 }
 
 /**
+ * CSS properties whose numeric value is unitless (aligned with React's
+ * `isUnitlessNumber` list, trimmed to properties actually reachable through
+ * plain `style.setProperty`). Every other property receives an implicit `px`
+ * suffix when given a bare number, matching what every mainstream framework
+ * (React, Vue, Preact) does for numeric style values.
+ * @type {Set<string>}
+ */
+const UNITLESS_NUMBER_PROPERTIES = new Set([
+	'animation-iteration-count',
+	'aspect-ratio',
+	'border-image-outset',
+	'border-image-slice',
+	'border-image-width',
+	'box-flex',
+	'box-flex-group',
+	'box-ordinal-group',
+	'column-count',
+	'columns',
+	'flex',
+	'flex-grow',
+	'flex-negative',
+	'flex-order',
+	'flex-positive',
+	'flex-shrink',
+	'font-weight',
+	'grid-area',
+	'grid-column',
+	'grid-column-end',
+	'grid-column-span',
+	'grid-column-start',
+	'grid-row',
+	'grid-row-end',
+	'grid-row-span',
+	'grid-row-start',
+	'line-clamp',
+	'line-height',
+	'opacity',
+	'order',
+	'orphans',
+	'tab-size',
+	'widows',
+	'z-index',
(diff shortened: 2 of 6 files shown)


chenzylab:
Thanks for the thorough review — all four issues were real gaps, fixed in the follow-up commits:

1. **Server omits implicit px suffix**: moved the allowlist + `format_style_value` into `client/utils.js` (already imported by the server entry for `is_ripple_object`), and had the server's `get_styles` route through it too. Added a dedicated server test.
2. **Custom properties receive implicit px**: `format_style_value` now leaves any `--`-prefixed property bare, mirroring `normalize_css_property_name`'s own `--` short-circuit.
3. **Prefixed unitless properties get px**: the allowlist is now expanded with `-webkit-`/`-moz-`/`-ms-`/`-o-` variants of every entry, the same way React expands vendor prefixes of its own list.
4. **Scale property incorrectly receives px suffix**: added the standalone CSS Transforms Level 2 `scale` property (distinct from the `scale()` function) — genuinely unitless, just newer than the React list this allowlist was seeded from.

Full client+server suite after all fixes: 112 files, 1355 passing, 4 skipped, 0 failing.

leonidaz:
Thanks for the contribution and for addressing the review feedback. After considering this further, I’m going to close this PR.

We’d prefer Ripple to keep CSS units explicit. A numeric value doesn’t communicate which unit the author intends, and defaulting to `px` introduces an implicit convention plus an allowlist we’d need to maintain. Solid, Vue, and Svelte also leave units explicit.

This `px` appending was introduced very early in React in v0.3.0 that was released on May 29, 2013, as its initial public release. It's an early API convention that doesn't apply everywhere, different theme values need different units. Ripple should preserve that choice by requiring explicit units.

Numbers should continue to work where CSS accepts unitless values, such as `opacity` and `line-height`. For dimensions, authors should specify the unit: `width: '400px'` or `width: '24rem'`.

The current `string | number` typing makes this confusing, so improving the types and documentation would be a better direction. Thanks again for highlighting that gap. I'll put up a pr for this shortly.

PR #1465, opened 2026-09-09
refactor(ripple): explicit TryState for try boundaries

### LINKED ISSUE #1464: refactor: explicit state for try boundaries (not a cold-compile win)
## Context

While optimizing cold hydration for `benchmarks/news` (#1463), Chrome tracing showed that a cold `mount()`/`hydrate()` is dominated by V8 lazily parsing and compiling the runtime functions it first calls. `try_block` in `packages/ripple/src/runtime/internal/client/try.js` stood out as the largest function on that path: about 10 KB unminified, with every helper (`begin_request`, `complete_request`, `handle_error`, `activate_streamed_chunk`, …) nested inside it as a closure created on every call.

## Correction to the original premise

The first version of this issue claimed the whole 10 KB was parsed and compiled on first call, including a re-preparse of the nested closures. That is wrong: V8 keeps preparse data for inner functions, so only `try_block`'s own outer body compiled; the closures compiled later, if ever. Splitting them out therefore moves bytes between functions without changing what compiles on the settled path. Measured on the news fixture under the default root boundary (#1465): `try_block` 10,220 → 1,331 B, functions compiled 71 → 68, cold hydrate unchanged within noise. The default boundary's remaining ~0.24 ms over `rootBoundary: false` is per-function overhead (about a dozen wrappers, closures and blocks on the settled path), not bytes.

No benchmark is affected either way: every fixture mounts with `rootBoundary: false`.

## What is still worth doing

Restructuring `try_block` around one explicit `TryState` object with module-level helpers (the shape `for.js` and `if.js` use) is a readability and allocation improvement: a boundary allocates its state and one branch closure instead of roughly fifteen closures. #1465 does that.

Any real cold-start gain for the default boundary would come from fewer functions on the settled path, not from shrinking this one.

leonidaz:
## Summary

`@try` blocks and the root boundary used to keep their state in ~20 mutable locals of one 10 KB `try_block` function, with every helper (pending, catch, request lifecycle, streaming activation) as a nested closure created on each call. This PR moves that state into one explicit `TryState` object (declared in `types.d.ts`) and makes the helpers module functions that take it.

After the follow-up commits the boundary interface is the state itself: `get_pending_boundary` / `get_boundary_with_catch` read `pending_fn` / `catch_fn`, the request, deferred and paused-block entry points pass `boundary.s` to the module functions, and `handle_boundary_error` replaces the `s.c(error)` calls in runtime.js. The try block and its branches are created with `block()` directly; `create_try_block` and `boundary_fn_running_block` are gone.

This is a structural cleanup, in the shape `for.js` (`ListState`) and `if.js` already use. It also removes per-boundary closure allocation: a boundary now allocates its state plus one branch closure, instead of roughly fifteen closures.

## What it is not

Not a cold-start win, despite the original motivation (#1464). V8 keeps preparse data for inner functions, so the old `try_block` never re-parsed its nested closures on first call; only its outer body compiled. Measured under the default root boundary on the news fixture:

| Measurement | Before | After |
| --- | ---: | ---: |
| `try_block` size, unminified | 10,220 B | 1,331 B |
| Source compiled during a cold hydrate | 35.3 KB | 26.0 KB |
| Functions compiled during a cold hydrate | 71 | 68 |
| Cold hydrate, interleaved fresh contexts | ≈ 2.5 ms | ≈ 2.5 ms (within noise) |

The default boundary still costs about 0.24 ms over `rootBoundary: false`; that is per-function overhead on the settled path, not bytes.

## Tests

- New client test: sibling `@try` boundaries resolve, fail and reset independently while keeping each other's DOM (also passes against the previous implementation).
- Client and hydration suites: 112 files, 1,315 tests passed, 12 skipped. Typecheck, formatting and `pnpm changeset:check` pass.

Closes #1464.

### DIFF
--- a/.changeset/try-boundary-cold-start.md
+++ b/.changeset/try-boundary-cold-start.md
@@ -1,0 +1,5 @@
+---
+'ripple': patch
+---
+
+Reduce cold mount and hydration compilation work for the default root boundary and `@try` blocks by moving pending, catch, and streaming helpers out of boundary initialization.
--- a/.changeset/try-boundary-cold-start.md
+++ b/.changeset/try-boundary-cold-start.md
@@ -2,4 +2,4 @@
 'ripple': patch
 ---
 
-Reduce cold mount and hydration compilation work for the default root boundary and `@try` blocks by moving pending, catch, and streaming helpers out of boundary initialization.
+`@try` blocks and the root boundary keep their state in one explicit `TryState` object, and the pending, catch, request and streaming helpers are module functions that take it. A boundary allocates its state and one branch closure instead of a closure per helper.
--- a/packages/ripple/src/runtime/internal/client/blocks.js
+++ b/packages/ripple/src/runtime/internal/client/blocks.js
@@ -217,8 +217,9 @@ export function root(fn) {
 }
 
 /**
- * @param {() => void} fn
- * @param {any} state
+ * @template T
+ * @param {(state: T) => void} fn
+ * @param {T} state
  * @returns {Block}
  */
 export function create_try_block(fn, state) {
--- a/packages/ripple/src/runtime/internal/client/blocks.js
+++ b/packages/ripple/src/runtime/internal/client/blocks.js
@@ -13,7 +13,6 @@ import {
 	TRY_BLOCK,
 	DETACHED_BLOCK,
 	HEAD_BLOCK,
-	DIRECT_CHILD_BLOCK,
 	IF_BLOCK,
 } from './constants.js';
 import { hydrating } from './hydration.js';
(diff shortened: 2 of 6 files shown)


leonidaz:
Reviewed and measured with the tooling from #1463 (fixture hydrated under the default root boundary by patching `rootBoundary: false` out of the built bundle).

**What the PR does, confirmed:** `try_block` 10,220 → 1,792 B in the unminified build; total source compiled during a cold hydrate 35.3 → 26.8 KB; function count unchanged (71).

**Why timing barely moves** (your table shows the same): V8 keeps preparse data for inner functions, so the old 10 KB `try_block` never re-parsed its nested closures on first call; only the outer body compiled. The byte reduction therefore converts to ~0.06 ms, and the remaining cost of the default boundary (~0.24 ms over `rootBoundary: false`, 2.28 vs 2.04 ms interleaved) is per-function overhead: about a dozen wrappers, closures and blocks on the settled path. Issue #1464 overstated the re-parse part; the structural cleanup is still the right shape.

**Pushed 99f7f8d7b on top:** the boundary state no longer carries the callback adapters (`b`/`r`/`c`/`rd`/`pb`/`rp`), which allocated six closures per boundary purely to forward to the state functions. The request, deferred, paused-block and error entry points now take `boundary.s` directly (`handle_boundary_error` replaces the `s.c(error)` calls in runtime.js), `TryState` is declared once in types.d.ts, and the try block and its branches are created with `block()` directly rather than through `create_try_block` → `boundary_fn_running_block` → `branch` (both wrappers removed). Coverage: 71 → 68 functions, `try_block` 1,331 B. Three-way interleaved A/B (main / PR / this) is at the noise floor on my loaded machine (~2.5 ms each), so treat this as a cleanup, not a win.

Client + hydration suites: 1315 passed; typecheck and format clean.

PR #1469, opened 2026-09-09
feat: custom transport for trackAsync hydration and RPC

### LINKED ISSUE #1467: Custom serializers (transport) for trackAsync hydration and RPC
## Summary

Ripple serializes `trackAsync` results (server → client hydration) and RPC arguments/results with [devalue](https://github.com/sveltejs/devalue), always with devalue's built-in types only. There is no way for an app to teach both sides about its own types — class instances, branded values, `Temporal`-like objects, ORM models, etc. devalue supports this through `reducers` (server, `stringify(value, reducers)`) and `revivers` (client, `parse(text, revivers)`); SvelteKit exposes the same idea as a `transport` config (`{ [name]: { encode(value) => data | false, decode(data) => value } }`).

Today a class instance in a `trackAsync` value throws `DevalueError: Cannot stringify arbitrary non-POJOs` during SSR.

## Proposed shape

`ripple.config.ts`:

```ts
export default defineConfig({
	transport: {
		Money: {
			encode: (value) => value instanceof Money && [value.amount, value.currency],
			decode: ([amount, currency]) => new Money(amount, currency),
		},
	},
});
```

`encode` returns `false`/`undefined` for values it does not own (devalue reducer semantics); `decode` rebuilds the value on the client. One definition, used by hydration and RPC in both directions.

## What has to change

Context: #1466 changed how `trackAsync` results are encoded, and a transport option has to fit around that.

**Runtime, server** (`packages/ripple/src/runtime/internal/server/index.js`)
- `serialize_track_async_result` now checks `is_plain_data(value)` and, when the value is a tree of strings / finite numbers / booleans / `null` / arrays / plain objects (each object reachable once, no symbol keys), emits the envelope as `{"ok":true,"value":<JSON.stringify(value)>}`. Only other values go through `devalue.stringify(value)` as `{"ok":true,"payload":"<devalue>"}`.
- **A reducer can claim a plain object** (`encode` may match on shape, e.g. `{ __type: 'Money', … }`), so whenever a transport is configured the JSON fast path must be skipped, or `is_plain_data` must additionally call every `encode` on every node and bail when one matches. Skipping is simpler and correct; the fast path is a pure optimization. Either way `devalue.stringify(value, reducers)` receives the configured reducers.
- `serialize_track_async_error` and the stream error envelope (`unit.error`) carry only a message string; no change.
- `packages/ripple/src/runtime/internal/server/rpc.js`: `devalue.parse(rpc_arguments_string, revivers)` for incoming args and `devalue.stringify({ value: result }, reducers)` for the result.

**Runtime, client** (`packages/ripple/src/runtime/internal/client/runtime.js`, `track_async`)
- The hydration read is `envelope.payload === undefined ? envelope.value : devalue.parse(envelope.payload)`; the `payload` branch becomes `devalue.parse(envelope.payload, revivers)`. The `value` branch stays as is (a transport-configured server never emits it for values a reducer could claim).
- `packages/ripple/src/runtime/internal/client/rpc.js`: `devalue.stringify(args, reducers)` outgoing, `devalue.parse(data, revivers).value` incoming.

**Plumbing**
- `RippleConfigOptions` in `packages/vite-plugin/types/index.d.ts` gains `transport`; `packages/vite-plugin/src/load-config.js` validates it (object of `{ encode, decode }` functions) next to `validate_root_boundary`.
- The config reaches the runtimes the way `rootBoundary` does: `packages/vite-plugin/src/server/virtual-entry.js` passes `rippleConfig.rootBoundary` into the server entry; the transport needs the same path on the server and an equivalent on the client entry, since `decode` must run in the browser bundle. Both runtimes need a place to hold the active transport: either an option on `render()` / `hydrate()` / `mount()`, or a `setTransport()` registration the generated entries call at startup (RPC calls happen outside any render, so a module-level registration is the simpler fit).
- The adapters (`packages/adapter-*`) that call `executeServerFunction` for RPC need the same transport available on the server.

**Tests and docs**
- Server: add transport cases to `packages/ripple/tests/server/track-async-serialization.test.tsrx` (`hydration envelope encoding` already asserts which values take the JSON path vs. the devalue path; with a transport configured, plain objects must take the devalue path).
- Hydration: `packages/ripple/tests/hydration/track-async-serialization.test.js` round-trips through the real client `track_async`; add a custom-type case and an RPC case.
- `website/public/llms.txt` and the config docs: document `transport`, and note that hydration payloads for plain data are raw JSON (no devalue) unless a transport is configured.

## Performance requirements (streaming-ssr must not regress)

#1466 got the `streaming-ssr` suite to where Ripple leads every op (all-fast total 1.11–1.17 ms vs Solid 1.30–1.36; cpu-100 1.24 ms vs octane 4.3; cpu-800 14.8 ms vs octane 30; cpu-waves-50 0.83 ms vs octane 7.5), largely by taking devalue off the per-card path. A transport must not put it back:

1. **No transport configured = today's code path, byte for byte.** With no `transport` in the config, `serialize_track_async_result` keeps the `is_plain_data` → `JSON.stringify` envelope, and the client keeps reading `envelope.value` without touching devalue. The `hydration envelope encoding` tests in `packages/ripple/tests/server/track-async-serialization.test.tsrx` pin which values take which path; they must keep passing unchanged.
2. **No per-node cost for an empty transport.** Whether a transport exists is decided once per serialization (or once at registration), never inside the per-value walk. No reducer/reviver object is built per call; the runtime holds one pre-built `reducers` and one `revivers` object created at registration and reuses them for every `trackAsync` result and every RPC call.
3. **No per-render lookups.** The active transport is a module-level binding set once by the generated server/client entry (the `rootBoundary` plumbing in `packages/vite-plugin/src/server/virtual-entry.js` shows the path). Nothing on the render or hydration path reads config, awaits an import, or walks a Map to find it.
4. **With a transport, one walk, not two.** Either skip the JSON fast path and let `devalue.stringify(value, reducers)` do the single pass, or fold the `encode` checks into `is_plain_data` so a value that contains no custom types still takes the JSON path. Do not walk the value with `is_plain_data`, bail, and then walk it again with devalue. Whichever is chosen, measure it: a plain card payload under a configured-but-unused transport must cost no more than one `devalue.stringify` today (~6.7 µs per card on this machine with the patched devalue).
5. **Client bundle stays flat for apps without a transport.** The `bundle-size` suite tracks `js_raw`/`js_gzip`/`js_brotli`; the transport wiring lives in the generated entry (or is otherwise tree-shakable) so an app with no `transport` ships the same runtime bytes as before. `decode` is only invoked on the `payload` branch of the hydration read; the `value` branch gains no work.
6. **Gate: run the suite before and after.** `node benchmarks/streaming-ssr/run.mjs 30` from the repo root (all six targets build now), plus `ssr-throughput` (`news-50/render`, `news-500/render`) and `bundle-size`. Report the ripple rows in the PR. Acceptance: `total_allfast`, `total_cpu_100`, `total_cpu_800`, `total_cpu_waves_50`, `news-*/render` within 3% of the pre-change run on the same machine (these are stable to ~1–2% now; `shell_*` and `total_staggered` are noisier and only need to stay ahead of the next target), `js_gzip` unchanged for the fixtures that configure no transport, and the correctness and browser swap gates passing.
7. **Add a fixture case, not a benchmark case.** A transport test belongs in the server/hydration test suites (round-trip a custom class through `trackAsync` and through RPC, in both directions). Do not add a transport to the `streaming-ssr` fixture's `data.ts`; it is byte-identical across the six targets and stays plain data.

## Notes

- Ripple is on devalue 5.9.2 (with the stringify performance patch from sveltejs/devalue#190 applied through pnpm `patchedDependencies`). devalue 5.9 also added an `operations` option for customizing how values are introspected; a transport does not need it.
- Naming: devalue says reducers/revivers, SvelteKit says `transport` with `encode`/`decode`. `transport` reads better in a config file; the runtime can map it to reducers/revivers internally.

leonidaz:
Custom `transport` handlers in `ripple.config.ts` now encode and decode application types for `trackAsync` hydration and RPC arguments/results. A `Money` instance retains its methods after SSR, hydration, and subsequent server calls.

The Vite plugin validates and registers handlers in both entries, including config reloads in development. `setTransport` supports custom integrations and dynamic/namespace imports. Includes public types, documentation, a patch changeset, and server, hydration, and Vite integration tests.

Closes #1467.

### Serialization and bundle behavior

Registration builds the reducer/reviver tables once and switches the server serializer once. The original default serializer remains unchanged. Configured hydration payloads embed devalue's flattened JSON directly and use `unflatten` on the client, avoiding a second string encoding and JSON parse. The client also accepts string devalue payloads. Error envelopes are unchanged.

The client build checks retained registration across the complete output graph before removing unused transport branches. This preserves manual and dynamic registration while keeping apps without a transport unchanged. Config helpers have a browser entry so importing `defineConfig` does not load the Vite plugin's Node.js dependencies.

### Validation

- `pnpm test`: **2,271 passed**, 12 skipped.
- `pnpm typecheck`, formatting, and changeset checks pass.
- Chromium: development and production hydrate a custom class without refetching; subsequent RPC arguments and results revive correctly.
- All streaming correctness and browser swap gates pass; bundle-size browser interaction gates pass.
- Tests cover class identity, shared references, shape-based encoders, single property reads, special devalue values, script escaping, empty/reset registration, legacy string payloads, and RPC in both directions.

### Performance requirements

1. **Default path preserved:** the streaming SSR and both buffered SSR production bundles are byte-for-byte identical to the pre-change builds. The existing raw JSON envelope tests remain unchanged and pass.
2. **Empty transport:** registration normalizes it to the default serializer; reducer/reviver tables are constructed only at registration.
3. **No render-time config lookup:** generated entries register at startup; the server switches the serialization function once, with no per-result transport lookup.
4. **One value traversal:** the configured path makes one direct `devalue.stringify(value, reducers)` call and never runs `is_plain_data`. Flattened JSON embedding reduces complete envelope cost below the original one-devalue envelope budget (measurements below).
5. **Client bytes unchanged:** all three benchmark fixtures retain identical raw/gzip/Brotli sizes. An additional real `trackAsync` hydration app without a transport also produces identical files and sizes.
6. **Regression gate:** streaming totals and paired buffered SSR results stay below the +3% regression limit. Every `shell_*` operation and `total_staggered` remains ahead of the next target. Correctness and browser gates pass.
7. **Fixtures:** custom transports are tested in correctness fixtures; benchmark sources and datasets are unchanged.

### Ripple timing rows

Baseline: `2631804cd`, Node 24.18.0, Intel Core i9-9980HK, macOS x64. Values are milliseconds; lower is better. Streaming uses the requested 30-render suite across all six targets. Buffered SSR uses medians of three paired, alternating-order runs at the standard 10-second budget per configuration.

| Operation | Before | After | Change |
| --- | ---: | ---: | ---: |
| `shell_staggered` | 0.13345 | 0.13572 | +1.71% |
| `total_staggered` | 50.10126 | 50.07843 | -0.05% |
| `shell_allfast` | 0.06482 | 0.07000 | +7.99% |
| `total_allfast` | 1.15173 | 1.12672 | -2.17% |
| `shell_cpu_10` | 0.06063 | 0.05663 | -6.61% |
| `total_cpu_10` | 0.16371 | 0.15961 | -2.51% |
| `shell_cpu_100` | 0.32842 | 0.37072 | +12.88% |
| `total_cpu_100` | 1.57134 | 1.49018 | -5.17% |
| `shell_cpu_800` | 6.25962 | 4.20065 | -32.89% |
| `total_cpu_800` | 19.53741 | 18.03291 | -7.70% |
| `shell_cpu_waves_50` | 0.15865 | 0.14944 | -5.81% |
| `total_cpu_waves_50` | 0.91689 | 0.89347 | -2.55% |
| `news-50/render` | 0.07392 | 0.07536 | +1.95% |
| `news-500/render` | 1.35432 | 1.15538 | -14.69% |

The earlier single-run table did not establish a regression: repeated executions of identical bundles varied in both directions. The paired buffered results above are the final comparison, with a maximum measured regression of **1.95%**. Apparent default-renderer speedups are timing variation, not an optimization claim.

All paired buffered samples (ms), included to show that variation:

| Pair | Version | news-50 | news-500 |
| --- | --- | ---: | ---: |
| 1 | before | 0.07392 | 1.35432 |
| 1 | after | 0.07536 | 1.14949 |
| 2 | after | 0.08193 | 1.17314 |
| 2 | before | 0.06889 | 1.37123 |
| 3 | before | 0.07636 | 1.25671 |
| 3 | after | 0.06861 | 1.15538 |

### Configured-but-unused transport cost

Plain cards from the unchanged streaming fixture, one prebuilt unused reducer, 20 measured batches of 10,000 cards after five warmup batches. The value encoder is called directly once per payload. Envelope timing includes framing and script escaping; the original-envelope control uses the pre-change serializer with its devalue path selected.

| Operation | Median µs/card |
| --- | ---: |
| Direct devalue, built-in types | 6.821 |
| Direct devalue, same prebuilt unused reducer | 7.169 |
| Original one-devalue hydration envelope | 8.867 |
| Configured transport hydration envelope | 7.562 |

### Client byte gates

| Fixture | Raw before / after | Gzip before / after | Brotli before / after |
| --- | ---: | ---: | ---: |
| keyed table | 38,525 / 38,525 | 14,214 / 14,214 | 12,739 / 12,739 |
| TodoMVC | 39,916 / 39,916 | 15,342 / 15,342 | 13,815 / 13,815 |
| chat-stream | 37,380 / 37,380 | 14,810 / 14,810 | 13,370 / 13,370 |
| Additional hydration app | 38,762 / 38,762 | 14,719 / 14,719 | 13,231 / 13,231 |

Identical server bundle SHA-256 values before and after:

- `streaming` (27,804 bytes): `0ff481865107aad3951a7e9694141bfab45a17474fcadb20977d237d863a6a28`
- `news-50` (54,299 bytes): `7d0e5c86f82c58316c4aed8e484600401fe16eb5de7843e3f2da94e1147d95ba`
- `news-500` (389,953 bytes): `518eb19500d5da29e7144f017a7c1e53426af11bab7189a58917027acbd2d55e`

Commands: `node benchmarks/streaming-ssr/run.mjs 30`, `BENCH_TARGETS=ripple CONFIGS=news node benchmarks/ssr-throughput/run.mjs`, and `BENCH_TARGETS=ripple node benchmarks/bundle-size/run.mjs`. Paired buffered runs reuse the saved before/after builds with `--no-build` and alternate their order.

---

> [!NOTE]
> **Medium Risk**
> Changes the serialization path for SSR hydration and RPC; mismatched or buggy handlers could break custom types, though apps without transport keep prior behavior.
> 
> **Overview**
> Adds optional **`transport`** handlers in `ripple.config.ts` so apps can round-trip custom types through **`trackAsync` SSR hydration** and **server RPC** using shared **`encode` / `decode`** pairs built on devalue reducers/revivers.
> 
> Runtime changes route client RPC and hydration revival through a small transport layer (`setTransport`, new `ripple/transport` export) instead of calling devalue directly everywhere. On the server, registering a nonempty transport swaps **`trackAsync`** serialization once to embed devalue’s **flattened payload** in the hydration script tag (client **`revive`** accepts flattened or legacy string payloads). Empty or missing transport keeps the existing plain-JSON hydration fast path.
> 
> The Vite plugin **validates** transport config, exposes a **browser-safe `defineConfig`** entry, and **auto-registers** transport in generated client/server entries (dev reloads config before RPC too). Tests and docs cover Money-style classes, shared refs, and manual registration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf4ffbf98671d6d8ee00a97fab9ebb8cf7f873b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leonidaz:
Reviewed the timing rows against the bundles. Built the streaming SSR fixture and the news SSR and client bundles with this branch's runtime and with main's (`2631804cd`) runtime on the same machine: SHA-256 identical in all three cases (`0ff48186…`, `7d0e5c86…`, `2b20fad7…`). Identical bytes cannot regress, so the `shell_*`, `news-50/render` and the larger "improvements" in the table are run-to-run variation, as the paired samples already suggested. Nothing to fix on the measured paths.

One adjustment pushed: dropped the `__RIPPLE_TRANSPORT__` define. It required a `renderChunk` hook that re-ran esbuild over every client chunk containing the token (a second parse/print of the runtime chunk with its own source map) plus a global and `typeof` checks at three call sites, and the whole mechanism bought about 65 bytes gzip in apps that use `trackAsync` or RPC, and nothing at all in apps that don't (the code is tree-shaken either way; the benchmark client bundles are byte-identical with or without it). The client now has two helpers in `internal/client/transport.js`: `encode(value)` for RPC arguments and `revive(payload)`, which takes either a devalue string or the flattened array a configured server embeds, and passes the registered revivers if any. `track_async` and `rpc` call those; no global, no chunk post-processing.

Server side is unchanged: the one-time serializer switch keeps the default path byte-identical, which the hashes above confirm.

PR #1488, opened 2026-09-17
perf: flat element templates built with DOM calls; force layout before timed benchmark samples

### LINKED ISSUE #1451: benchmarks: force layout before timed samples in the remaining browser harnesses
## Summary

Browser benchmark harnesses time a DOM mutation right after `gc()`, but leave it to the browser whether the existing DOM has been laid out when the timer starts. A table whose rows were never laid out reorders about 2x faster than one with layout boxes, so samples split into two modes depending on whether a frame happened to render during the settle sleep. This produced misleading competitor numbers in `js-framework-reorder` and can affect every suite that mutates an existing tree.

The `js-framework` harnesses were fixed in #1450 (`benchmarks/js-framework/click.mjs` and `run-reorder.mjs`) by forcing style and layout before the timer:

```js
(window.gc || (() => {}))();
void document.body?.offsetHeight;
const t0 = performance.now();
```

The other browser harnesses still have the nondeterminism.

## Evidence

Raw `reverse` samples from three independent normal runs of `js-framework-reorder` before the fix (ms):

```
solid  run1  1.82 1.75 1.82 3.88 3.13 3.20 3.72 3.78
solid  run2  3.15 1.80 1.75 1.83 3.15 3.13 1.75 3.10
solid  run3  3.10 3.10 3.13 3.08 3.10 3.18 1.70 1.72
ripple run1  2.55 2.68 2.63 2.57 3.00 2.55 2.60 2.63
```

Solid's samples alternate between a ~1.7 ms mode and a ~3.1 ms mode; Ripple's are all in one mode. Which mode a sample lands in depends on whether the browser rendered a frame between the harness's `#run` reset and the timed clicks. With layout forced, Solid measures 2.94 ms consistently and the bimodality disappears (recorded baseline: Ripple 2.93, Solid 2.94).

## Affected harnesses

Each of these has one or more `window.gc` → `performance.now()` timing sites without a preceding layout flush:

- `benchmarks/chat-stream/run.mjs`
- `benchmarks/dbmon/run.mjs`
- `benchmarks/effectful-list/run.mjs`
- `benchmarks/memo-wall/run.mjs`
- `benchmarks/portal-swarm/run.mjs`
- `benchmarks/recursive-context/run.mjs`
- `benchmarks/signal-favoring/run.mjs`
- `benchmarks/todomvc/run.mjs`
- `benchmarks/uibench/run.mjs`
- `benchmarks/news/run.mjs` (hydrate)
- `benchmarks/js-framework/run-anchors.mjs`

Ops that start from an empty container are unaffected in practice, but the read is harmless there and applying it uniformly is simpler than deciding per op.

## Proposal

Add `void document.body?.offsetHeight;` between the `gc()` call and `performance.now()` at every timing site (the optional chain keeps the harness unit tests, which run against a fake document, working). Because the harness files are part of the workload fingerprint, each suite's baseline must be re-recorded afterwards; doing it as the first step of each suite's optimization pass keeps every recording consistent with its harness without a full-matrix re-record up front.

This matches what the upstream krausest benchmark measures, where paint is inside the timing window and rows are always laid out.

leonidaz:
## Summary

**Flat element templates.** A template that is exactly one element with static attributes and at most one text child now compiles to `_$_.template_el(tag, attrs, text)` and is built with `createElement` / `setAttribute` / `textContent` instead of a `<template>.innerHTML` parse. Blink's fragment parser has a fixed cost per call that dwarfs such an element (~7.5 µs vs ~0.75 µs cold), and an app's first render pays it once per distinct template. Raw-text elements, `pre`/`textarea`, `svg`/`math` roots, custom elements, `is` attributes, fragments and anything with a placeholder or dynamic part still parse. Hydration is unchanged (`H.t(false, 1)`).

**Client-only builds** no longer emit a cursor-only sibling read that Rollup kept as a dead `div.firstChild;` statement.

**Benchmark harnesses force layout before every timed sample** (closes #1451). Removing or moving nodes that own layout boxes costs up to several times more than on a tree the browser has not laid out yet, and a fixed settle sleep hit that state in a random third of samples, so unmount scores were bimodal. Every browser timing site now has `void document.body?.offsetHeight;` between `gc()` and `performance.now()`, matching what #1450 did for `click.mjs` / `run-reorder.mjs`: chat-stream, dbmon, effectful-list, memo-wall, portal-swarm, recursive-context, signal-favoring, todomvc, uibench, news (hydrate) and js-framework/run-anchors. Documented in `benchmarks/README.md`.

**Fixtures.** Every Ripple fixture gets dbmon's `tsconfig.json` and `textTypes: { tsconfig: 'tsconfig.json' }` (the news fixture's hand-rolled plugin uses `createTextTypeProject`, like streaming-ssr). The signal-favoring fixture builds with `ssr: false` like the other client-only fixtures.

## Results (signal-favoring, official runner, 20 iterations, same machine)

| op | Ripple before | Ripple after | Inferno | Vue Vapor |
| --- | --- | --- | --- | --- |
| mount | 4.10 ms | 2.76–2.80 ms | 3.20–3.31 ms | 5.7 ms |
| unmount (laid-out tree) | lottery, 0.075–0.087 | 0.250 | 0.250–0.313 | 0.275–0.288 |

Ripple already led every bump op. With fine-grained timers, Ripple's teardown is ~3 µs above the raw DOM-removal floor (0.023 ms vs Inferno 0.058 unsettled; 0.24 vs 0.27 after layout). recursive-context unmount under the new convention: Ripple 1.20 vs Inferno 1.44; dbmon 5.06 vs Inferno 6.08 / Preact 6.32.

## Methodology audit

All nine signal-favoring adapters mount synchronously in the timed window on a fresh page, flush every bump inside the call, render the same 100-component chain and pass the same semantic gate; only whitespace text nodes differ, which the gate normalizes.

## Follow-ups

- Harness files are part of the workload fingerprint, so the baselines for the touched suites need re-recording (planned for the coming days); the unmount rows in the committed baseline report are stale until then.
- async-waterfall, effectful-list, memo-wall, portal-swarm, recursive-context, dbmon and uibench are client-only apps that still build with hydration support; `ssr: false` there would be consistent but was left out of this PR.

## Validation

`pnpm test` (2278 passed), `pnpm typecheck`, `pnpm format:check`, `pnpm changeset:check`, the benchmark package tests, and quick gate runs of all eleven touched suites for Ripple.

---

> [!NOTE]
> **Medium Risk**
> It changes a core first-render path for many components and tightens benchmark methodology, so regressions could show up in hydration, namespace edge cases, or recorded baselines—not in security-sensitive areas.
> 
> **Overview**
> **Flat element templates.** The client compiler now recognizes static templates that are a single HTML element (static attrs, at most one text child) and emits **`template_el`** instead of parsing a `<template>` string. The runtime builds those nodes with **`createElement` / `setAttribute` / `textContent`** in an inert document and clones them like parsed templates; fragments, SVG/MathML, table-model foster-parenting cases, custom elements, `is`, and anything dynamic still use **`template`**. Hydration stays on **`H.t(false, 1)`**.
> 
> **Client-only codegen.** When appending into a parent without a placeholder, sibling cursor reads for hydration are skipped if the build is not hydrating—avoiding dead **`firstChild`** reads after bundling.
> 
> **Benchmarks.** Browser harnesses insert **`void document.body?.offsetHeight`** after **`gc()`** and before **`performance.now()`** so timed work starts from a laid-out tree (documented in **`benchmarks/README.md`** and uibench). Ripple benchmark apps gain local **`tsconfig.json`** files and **`textTypes: { tsconfig: '…' }`** on the Vite plugin; the news bench wires **`createTextTypeProject`** into its custom compile hook.
> 
> Tests and golden compiled output were updated to expect **`template_el(...)`** where applicable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30dacd2d96ab8eff4355dee220beec0f16a4c38c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leonidaz:
Addressed both Bugbot findings in the latest commit, with a different fix for the first one:

- **Media prefetch:** `template_el` now creates its cached master in the inert template document (`document.createElement('template').content.ownerDocument`), the same document that owns parsed template content. A node there has no browsing context, so `src`/`srcset`/`poster` do not load until a clone is adopted into the page. Verified in Chromium: an `img`, `video` and `input type=image` created there fetched nothing at creation and each clone fetched exactly once when appended, identical to a parsed `<template>`. This keeps `` templates on the fast path instead of deny-listing attributes; Firefox imports the node like `template` does.
- **Table-model text:** non-whitespace text directly inside `table`/`thead`/`tbody`/`tfoot`/`tr`/`colgroup` is foster-parented by the parser, so those templates now keep parsing. Whitespace-only text in those elements and text inside `select` are inserted as written by the parser, so they stay flat (covered by tests).

Cold mount on signal-favoring is unchanged (2.6 ms vs Inferno 3.2). Full test suite, typecheck and format check pass.

PR #1493, opened 2026-09-17
fix(adapter-bun): serve a prerendered directory's index.html

Dev-next-gen:
#1487 taught the Node and Bun adapters to serve a prerendered page's `<path>/index.html` before the server renders. On Node that works, but on Bun the new branch never runs. The Bun middleware assumed `Bun.file(dir)` reports `exists() === true` with `size === 0`. Real Bun reports the opposite: `exists()` is false for a directory, and `size` is the directory's size on disk. So `/about` returned before reaching the `index.html` lookup and fell through to the server render. The existing test mock also returned `exists: false` for directories, and no test requested a directory, so nothing caught this.

I checked this with a small script that calls `serveStatic()` on a folder containing `about/index.html`, under Bun 1.3.10 (the version in the lockfile), 1.3.11 and 1.4.2. All three print `exists(): false size: 4096` for the directory. Before this change, `/about`, `/about/` and `/` all fell through to `next()`. After it, they return 200 with `text/html; charset=utf-8` and the page body. `/missing` and `/missing/` still fall through.

The fix tries `<path>/index.html` when the requested path doesn't exist, and reassigns `file_path` so the content type comes from the `index.html` file, as in the Node adapter. The previous code would have used the directory path here and sent `application/octet-stream` if the branch had ever run. I dropped the `size === 0` check because its premise doesn't hold. The only other behavior change is that a zero-byte file such as an empty `robots.txt` is now served with a 200 instead of falling through, which also matches Node. The test mock now reports directories the same way Bun does, and a new test requests a prerendered directory. That test fails on `main` (the middleware calls `next()`) and passes with the fix. `pnpm vitest run --project adapter-bun` passes (15 tests), and so do `tsgo` on `packages/adapter-bun/tsconfig.typecheck.json` and `prettier --check`. I added a patch changeset for `@ripple-ts/adapter-bun`.

---

> [!NOTE]
> **Low Risk**
> Scoped change to Bun static file serving and tests; no auth, data, or core runtime changes.
> 
> **Overview**
> Fixes **Bun static middleware** so prerendered routes (e.g. `/about` with `about/index.html` on disk) are served instead of falling through to SSR.
> 
> Bun reports **`Bun.file(directory).exists() === false`**, so the old logic that treated directories as existing with **`size === 0`** never ran the **`index.html`** lookup. When the resolved path is missing, the middleware now tries **`<path>/index.html`**, updates **`file_path`** for MIME headers, and only then calls **`next()`**. **Zero-byte** static files (e.g. empty **`robots.txt`**) are served with **200** instead of falling through, matching the Node adapter.
> 
> Tests update the **`Bun.file`** mock to mirror real Bun directory behavior and add coverage for prerendered index, missing index fallthrough, and empty files. Patch changeset for **`@ripple-ts/adapter-bun`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7fdc247cf81468efe0db2a6e4ddfc11d3c4aace. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leonidaz (approved):
thank you!

PR #1494, opened 2026-09-17
fix(adapter): apply init when a same-origin Request is short-circuited

Dev-next-gen:
While reading `patch_global_fetch` in `packages/adapter/src/rpc.js`, I noticed that the same-origin short-circuit only uses `init` when the input is a string or a URL:

```js
const request = input instanceof Request ? input : new Request(input, init);
```

When server code calls `fetch(request, init)` with a `Request` for its own origin, the handler receives the original `Request` and `init` is ignored. A real `fetch` applies `init` over the request, so the same call behaves differently depending on whether it goes out to the network or stays in the process.

To check it, I added `packages/adapter/tests/rpc.test.js`, which patches fetch, sets a handler and calls `fetch(new Request('http://localhost:3000/api/items'), { method: 'POST', body: 'payload', headers: { 'x-test': '1' } })`. On `main` the handler gets `{ method: 'GET', body: '', header: null }`. With the change it gets `{ method: 'POST', body: 'payload', header: '1' }`, and `vitest run --project adapter` goes from 1 failed / 10 passed to 11 passed.

The fix builds the request with `new Request(input, init)` in every case, which is what `fetch` does. When `init` is undefined this makes a copy of the `Request` with the same method, headers, body and signal, so calls without `init` behave as before. I added a patch changeset for `@ripple-ts/adapter`. `tsgo --noEmit -p packages/adapter/tsconfig.typecheck.json` and prettier pass.

---

> [!NOTE]
> **Medium Risk**
> Changes server-side global fetch patching used during SSR and same-origin handler routing; incorrect behavior could have affected route handlers, but the fix aligns with the Fetch spec.
> 
> **Overview**
> Fixes **same-origin in-process `fetch` short-circuiting** so it matches native `fetch` when callers pass `fetch(request, init)`.
> 
> The handler path now always builds the request with **`new Request(input, init)`**, so method, body, headers, and signal from `init` override the original `Request` instead of being dropped (e.g. a `POST` no longer arrives as the original `GET`).
> 
> **Error handling** is tightened to mirror native `fetch`: synchronous failures during URL rewriting, `Request` construction, or handler invocation **`Promise.reject`** instead of throwing or silently falling back to a real network request. Invalid relative URLs and bad `init` no longer hit the network when the short-circuit path would apply.
> 
> Adds **`packages/adapter/tests/rpc.test.js`** covering init merging, promise identity, and network passthrough for cross-origin / missing context / no handler, plus a patch **changeset** for `@ripple-ts/adapter`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 268dbe94a9986348fca757a217471fa244a705ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leonidaz (approved):
thanks! i also adjusted our patched fetch handler.

PR #1499, opened 2026-09-18
fix(rollup-plugin): pass compilerOptions to compile() instead of file id

Dev-next-gen:
The Rollup plugin passed the file path as the third argument to `compile()`, which expects a `CompileOptions` object. With the current compiler, even a default `.tsrx` build throws `TypeError: Cannot use 'in' operator to search for 'textTypeFacts'` because that argument is a string.

Forward the existing `compilerOptions` object so builds succeed and options such as `mode: 'server'`, `dev`, `hmr`, and `hydration` take effect. Include a patch changeset and register a Rollup Vitest project with regression tests against the real compiler, including client/server Rollup builds, option behavior, CSS extraction, and source maps.

Document the Vite/compiler option boundary: `rootBoundary` controls bundled runtime code, `excludeRippleExternalModules` controls package scanning, and `textTypes` manages the TypeScript project that produces per-file `textTypeFacts`. These remain Vite plugin options; Vite already derives and forwards its compiler options correctly.

Validation:

- `pnpm test --project rollup-plugin --project vite-plugin`: 58 tests passed.
- Reverting the one-line compiler call reproduces the `TypeError` in all 10 new Rollup tests; restoring it makes them pass.
- `pnpm typecheck` passed.
- Prettier checks for changed files, `pnpm changeset:check`, and `git diff --check` passed.


leonidaz (approved):
Thank you!

